PR #2, opened 2026-08-17
Keep a pasted block as one message

riesbri:
Priority 2. Behaviour change, isolated to input handling; nothing about rendering or the session loop moves.

## The bug

A pasted paragraph fired one message per line. The decoder reported every embedded newline as `enter`, and the composer sends on `enter`.

## The fix

**Bracketed paste**, which is the only reliable signal — without it a pasted newline is indistinguishable from a pressed one. The terminal owner enables it on acquire (`?2004h`) and disables it on close (`?2004l`), and pasted content arrives as its own key kind whose newlines are literal.

**Decoding is now stateful.** A paste can be split across reads, and so can an escape sequence, so the decoder keeps the undecidable tail of a chunk and resumes on the next. Two cases that used to be lumped together are now distinguished:

- an **incomplete** sequence is held — terminals emit those in one burst, so a truncated one means the read split it, and reporting `escape` there dropped the real key;
- a **lone trailing** escape is still reported immediately — waiting would make Escape indistinguishable from a slow arrow key, and a spurious cancel is recoverable where a swallowed one is not.

**`alt-enter` inserts a newline without sending.** `shift-enter` is deliberately not bound: terminals send a bare carriage return for it, identical to `enter`. That is why the plan's "shift-enter / alt-enter" is only half implementable.

**The composer holds logical lines**, so it reports the cursor as a line plus a column, and the view maps that to a rendered row by wrapping each preceding line with its own gutter — a preceding line may itself have wrapped, so the cursor cannot be placed by counting characters.

## Verification

105 tests. The new ones cover a paste split across three reads, a chunk ending mid-terminator, typing before and after a paste staying separate, a control byte inside a paste treated as content, a sequence split across two reads, and the cursor reported per logical line.

Checked in a real terminal too — a three-line paste and an alt-enter newline both sit in the composer as one message, with continuation lines aligned under the prompt:

```
╭─ my-project ─────────────────────╮
│ › first line                     │
│   second line                    │
│   third line                     │
╰──────────────────────────────────╯
  ● ready · … · alt-enter newline · …
```

## Note

The plan's test was "paste a 5-line block, assert one `user/message` event". That assertion needs the harness and an API key, so the equivalent here is hermetic: assert the composer yields one `submit` carrying all the lines. The live check above covers the rendered half.

Adding the `paste` kind also surfaced an unhandled case in the select overlay, which typechecking caught — it now ignores non-key input rather than falling through.


riesbri:
Good catch, and correct — thank you. This was a real hole: the README states that everything a model, tool, or session log produces is untrusted for terminal purposes, and I had overlooked that **pasted input is untrusted too**, and is in fact the most likely source of terminal controls in the whole interface, because people paste logs.

Fixed in `9d3d5ab`. Sanitizing at the **point of insertion** rather than at render time, so the buffer holds one representation — otherwise every width, cursor, and draw calculation would read different text than the terminal receives, which is a second bug waiting to happen.

Three problems arrived together, so all three are handled:

- **Line endings** normalize to `\n`. As you noted, `Composer.lines` splits on `\n` alone, so a CRLF paste kept the `\r` inside every line — and a surviving `\r` returns the cursor to column zero.
- **Tabs** expand to spaces. A tab's rendered width depends on tab stops the width arithmetic cannot see, so measuring one is guesswork.
- **Everything else that is a control character** becomes caret notation via the existing `escapeControls`.

Typed text is deliberately left alone: control bytes arrive from the decoder as named keys, so only a paste can carry them.

Verified live in a pseudo-terminal with a hostile payload — a clear-screen sequence plus CRLF endings:

```
╭─ deepseek-harness ───────────────────────────╮
│ › ^[[2Jred^[[31m                             │
│   second line                                │
╰──────────────────────────────────────────────╯
```

The banner above it survives, which it would not have before, and the block still splits into logical lines. Four unit tests cover the escape sequence, CRLF and lone CR, tab expansion, and typed text being untouched.

PR #8, opened 2026-08-17
Tool cards from render intent, and a hanging indent for wrapped rows

riesbri:
Stacked on #7.

## Wrapped rows lost their gutter

Found by reading a 96-column pseudo-terminal capture rather than a test. A marked line is a mark followed by content, so it carries no leading whitespace for `wrapToWidth` to preserve, and every row after the first landed back at column zero. Model prose is one long logical line, so in practice **every paragraph of every reply was ragged at any real terminal width**:

```
● The instruction says a sentence with bold and code, so I need a sentence
containing bold text and inline code. Then a two-item bullet list.
```

`hangingIndent` wraps to the budget the gutter leaves and prefixes each row itself, so the rows it returns already fit and nothing wraps them again. It measures the indent in display columns, because a wide mark would otherwise let each row overflow by exactly the columns the glyph adds.

## Tool cards

`presentCall` and `presentResult` are pure functions of the call's arguments — the harness's contract — so a frontend may call them freely. All six result cards and all three call cards are handled:

| intent | drawn as |
| --- | --- |
| `terminal` | framed command headed by its working directory, output framed with its exit status |
| `diff` | red removals, green additions, per file |
| `search` | matches grouped under each file, or a flat path list, with a capped marker |
| `read` | the file's own line numbers, and `N of M lines` |
| `web` | sources with their urls |
| `generic` | title, category icon, and the result content |

A tool that declares nothing still renders — every intent is documented to degrade to raw content — and no card is ever invented for a tool by name.

### Two things the harness's card model assumes an appending screen cannot do

**A mutation tool returns its diff from both presenters on purpose**, because a completed card is meant to *replace* the pending one. Drawing both printed every change twice, once proposed and once applied. The change is now drawn once, at result time, which is also the more truthful of the two: it is what landed. A tool that declares a call-time diff and no result-time one keeps the proposal, so nothing is lost.

**`rawInput` is a detail-view field.** `grep`'s title already names the pattern its `rawInput` repeats, so every search printed `MIT` twice. It is now shown only at `full`.

### ctrl-o

Cycles `compact` → `full` → `hidden`, shown in the status line. Because finished output goes into the terminal's own scroll buffer and is never rewritten, it sets the level for cards drawn from then on rather than reflowing history — the trade for keeping native scrollback, selection, and copy, and stated as such in the README. `hidden` still draws the call and still reports a non-zero exit, because a transcript that omitted those would lie about what ran.

## Verification

- 218 unit tests, 36 of them for cards, including two through `@xterm/headless` asserting a terminal card's two frames land in the same columns and that nothing overflows a 30-column terminal.
- Ten mutations applied, each confirmed to fail a test that names the behaviour.
- Real pseudo-terminal runs against the live API exercising `bash` (with a non-zero exit), `read`, `grep`, `write`, `edit`, and the `ctrl-o` cycle.


riesbri:
Landed in #12, which was stacked on this branch and merged with its commits included — `feat/tool-cards`'s base (`feat/incremental-streaming`) was deleted when #7 merged, leaving this PR unable to compute mergeability.

Everything reviewed here is on `main`: the render-intent cards, the hanging indent, and all eight findings raised on this PR, each fixed and mutation-checked. Closing as superseded rather than merged.

PR #64, opened 2026-08-22
feat(tui): say what the goal is, and what a long turn is doing

riesbri:
> **Stacked on #63.** Base is `feat/tool-output-reachability`; both touch `views.ts`. Retarget to `main` once #63 lands.

## What a person saw

```
  ⠙ working 14m 26s · x-preview-f-free · ↑2.3M ↓21k · ▌░░░░░░░ 68k/1.0M · goal 0/256 · todo 5/11 · ctrl-c interrupt
```

Two questions: *why is there a goal — I never ran `/goal`?* and *why has it been 0/256 for fourteen minutes?*

## The investigation

Not a rendering bug. `ctx.goals.get(agent)` returns `undefined` when no goal exists (`dsh-goal/lib/index.js:796-810`), so the segment was telling the truth. The goal was created by the **model**: the harness publishes `create_goal` as a model-callable tool (`@deepseek-ai/dsh-tool-goal`, mounted by `@deepseek-ai/dsh-base`) whose own description reads

> "Create one persisted same-session completion goal when the current direct human request is a long-running objective… **You may infer that intent without requiring the user to say 'create a goal'.**"

So a session can acquire automatic continuation authority nobody typed, and this segment is the only standing signal that it did. That makes it the wrong thing to have been showing: `256` is the deployment's round cap (`defaultMaxGoalRounds`), `0` is `roundsStarted`, and the objective — the first question about a goal you did not set — was nowhere.

## The change

| Before | After |
| --- | --- |
| `goal 0/256` | `goal armed · ship the release` |
| `goal 3/256` | `goal 3/256 · ship the release` |
| `goal 3/256 idle` | `goal idle · ship the release` |
| `goal paused` | `goal paused · ship the release` |

The count appears only once a round has been taken; before that `armed` says the same thing in words that are true, where `0/256` reads as a meter stuck at zero.

## Keeping AGENTS.md §9 intact

An objective is prose, and prose does not fit *"a mode is dropped whole rather than shortened"*. Hence two labels rather than one: dropping the objective leaves `goal 3/256`, which is still true, where shortening `3/256` would be a different number. That is one extra rung in the `tails` ladder, placed so the objective goes before `plan` and long before the goal itself. Bounding the objective's own width happens in `modes.ts`, not `views.ts`, so the composition function still has no way to cut anything.

## Also: what the fourteen minutes were spent on

```
  ⠙ working 14m 26s · run_shell_command · x-preview-f-free · …
```

`ToolCards` already knows — `pending` holds every call awaiting a result — so `inFlight()` returns the newest and it sits beside the timer. Elapsed time alone reads identically whether a command is running or the session has hung. It is the first fact surrendered when the line runs out, being a convenience reading like `todo` and `work`.

## Tests

Both the objective and the tool name are untrusted text reaching the terminal on every frame; both are escaped and both have a test. `status-frames.spec.ts` is new — it resizes across thirteen widths on a real emulator and asserts nothing ever survives in half.

Broken on purpose, each separately:

| Reverted | Caught by |
| --- | --- |
| `armed` → `0/256`, objective dropped | `says armed rather than nought of a cap no round has approached` |
| the objective ladder rung | `gives up a goal's objective before the goal itself` |
| `activity` from the body ladder | `names the tool a long turn is waiting on` |

`pnpm build && pnpm typecheck && pnpm test` — 972 passed, 1 skipped. Changeset: `minor`.

Docs: `docs/usage.md` gains the new goal wording, a **While a turn is running** section, and an explicit note that a goal can start without you; the `/goal` warning callout and the `ROADMAP.md` limitation both say so too.

### DIFF
--- a/.changeset/lazy-cats-wonder.md
+++ b/.changeset/lazy-cats-wonder.md
@@ -1,0 +1,23 @@
+---
+"@riesbri/dsh-tui": minor
+---
+
+Say what the goal is, and what a long turn is doing.
+
+The goal segment read `goal 0/256` — a round cap the reader never chose, against a
+count that had not moved, for an objective it never named. It now leads with the
+objective (`goal armed · ship the release`) and reports the count only once a round
+has actually been taken. This matters because a goal is not always something the user
+set: the harness publishes `create_goal` as a model-callable tool and tells the model
+it may infer a long-running objective without being asked, so a session can acquire
+automatic continuation authority that was never typed. The status line is where that
+becomes visible, and it now says what it is.
+
+The objective is the one part of a mode that may be surrendered on its own as the
+terminal narrows — it is prose, so a shorter one is still true, where a shortened
+round count would be a different number. Everything else about the drop order is
+unchanged.
+
+The working segment also names the tool the turn is waiting on
+(`⠙ working 14m 26s · run_shell_command`). Elapsed time alone reads the same whether a
+command is running or the session has hung.
--- a/.changeset/lazy-cats-wonder.md
+++ b/.changeset/lazy-cats-wonder.md
@@ -19,5 +19,8 @@ round count would be a different number. Everything else about the drop order is
 unchanged.
 
 The working segment also names the tool the turn is waiting on
-(`⠙ working 14m 26s · run_shell_command`). Elapsed time alone reads the same whether a
-command is running or the session has hung.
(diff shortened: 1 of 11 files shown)


riesbri:
Closed automatically by GitHub when the base branch `feat/tool-output-reachability` was deleted on merging #63. Continued in #66 — same branch and commits, based on `main`.

PR #68, opened 2026-08-23
fix(tui): accept the published Harness 0.1.1-rc line in the peer ranges

riesbri:
## Summary

An audit against all three surfaces — the locked rc.8 line, the published `0.1.1-rc.2` line, and upstream `master` via source-link — found **no API drift**, while the peer ranges (`^0.1.0-rc.7`) rejected every `0.1.1-rc.x` package under npm's prerelease matching rules: metadata lying quietly, producing unmet-peer warnings / ERESOLVE errors for consumers even though the bundle ran fine. A packed plugin booted correctly beside the published launcher.

This PR fixes the metadata and replaces the weekly master-only probe with a continuous compatibility system.

## The compatibility system (final state)

**Daily workflow** (`.github/workflows/harness-compatibility.yml`), three jobs:

1. **`typecheck` — harness `master`.** Builds upstream declarations in a separate read-only checkout, links them with `tools/link-harness.mjs`, typechecks this project. Early warning for unreleased change.
2. **`released` — full suite on the published line.** `tools/sync-harness.mjs` first pins *every* `@deepseek-ai/*` devDependency in **both** manifests (bundle + workspace root) to the exact authoritative published version — whole-graph, no mixed old/new line possible. Then the peer-currency check records its verdict, and install → build → typecheck → full test suite → consumer smoke all run regardless. Final verdict prints both axes:
   ```
   runtime compatibility: PASS/FAIL
   peer contract: ACCEPTED/REJECTED
   ```
   so a newly published tuple rejected by the ranges still yields a complete runtime answer ("works, awaiting an explicit range decision") instead of an unexplained red.
3. **`open-update-pr` — one rolling sync PR.** When the published line moves, maintains a single branch `automation/harness-sync` (reset + force-push; open PR edited in place, created if none). Classification by the checker's exit code: green → "chore: track published Harness line"; rejected → title says a peer compatibility decision is required, body carries the warning, CI red by design.

**Peer contract:** ranges are `^0.1.0-rc.7 || ^0.1.1-rc.2` — both lines the suite has actually passed against, nothing newer. `pnpm run check-peers` verifies each range accepts its authoritative dist-tag target; **no automation ever widens a range** (sync physically cannot write `peerDependencies`; invariant is unit-tested). Widening stays an explicit human decision.

**Consumer smoke:** `pnpm run test:consumer` packs the bundle, installs the published launcher into a scratch prefix, adds the plugin to a fresh profile through `dsh plugin add`, boots under a pseudo-terminal, requires banner + readiness evidence, exits cleanly on ctrl-d.

## Security model

- Every job that executes Harness/dependency code holds `contents: read` and receives no secrets.
- The PR job's write token (`SYNC_TOKEN`, fine-grained PAT — same pattern as `VERSION_TOKEN`) exists only inside the single step doing git/gh plumbing; that step runs no dependency code (manifest rewrites, `--lockfile-only` resolution, git/gh only). Git authenticates via `gh auth setup-git`; the token never touches a command line or repo config.
- The 24-hour `minimumReleaseAge` supply-chain brake applies everywhere.

## Validation

999 tests passing (incl. 46 new across sync-harness / peer-currency / smoke-parser specs), build + typecheck clean at every dependency level tested, zizmor clean, live `sync-harness --check` green after pinning, live consumer smoke passed end-to-end on real registry versions. Deliberate-mistake checks recorded in commit bodies.

## Setup note

Automated sync PRs need the repository secret `SYNC_TOKEN` (Contents: write + Pull requests: write) because GITHUB_TOKEN-created PRs do not trigger CI. Without it, the daily job reports what moved and how to sync by hand instead of failing.

Co-Authored-By: ox-alpha <noreply@deepseek.ai>


riesbri:
## Follow-up pushed (`7a15a83`): tightened for fast-moving Harness releases

Addressing all six review points, surgically:

**1. Whole-graph released job.** `tools/sync-harness.mjs` (new) now pins *every* `@deepseek-ai/*` devDependency in **both** manifests — bundle and workspace root — to the exact authoritative published version before anything installs. No floating tags, no stale root pins, so a mixed old/new Harness graph is impossible by construction. The job's first step is the sync.

**2. Automated consumer smoke.** `tools/consumer-smoke.mjs` (+ spec) commits the audit's manual sequence: install published launcher → pack bundle → add to fresh profile via `dsh plugin add` → boot under a pseudo-terminal → require banner + readiness evidence → ctrl-d → require exit 0. Evidence matching runs on an escape-stripped, whitespace-free flattening because the renderer shatters words across writes. Ran for real on this machine against `@deepseek-ai/dsh@0.1.1-rc.2`: passed end-to-end.

**3. Daily cadence.** Cron is now `17 6 * * *`; the 24-hour `minimumReleaseAge` brake in `pnpm-workspace.yaml` is untouched and still governs every install.

**4. `sync-harness` helper.** Deterministic, zero dependencies, pure logic under test: reads each package's authoritative dist-tag (map lives once in `check-peer-currency.mjs`, imported by `link-harness` too — one definition of "current"), detects drift including tag→pin conversion, writes sorted manifests, `--check` mode exits 1 with a from→to table. A new sibling workflow job opens an automated sync PR when the line moves: branch `harness-sync/YYYYMMDD`, lockfile refreshed via `pnpm install --lockfile-only`.

**5. No blind peer expansion.** `sync-harness` physically cannot touch `peerDependencies` (only devDependencies keys are considered; invariant covered by a named test). The sync job classifies via the checker's exit code: green → routine-bump PR title; rejected → PR titled "peer compatibility decision required" with a warning block and CI expected red until a human decides.

**6. Trust boundaries preserved.** Both executing jobs keep `contents: read`. The PR-opening job holds scoped write but by construction executes no dependency code — manifest rewrites, `--lockfile-only` resolution (no linking, no lifecycle scripts), git/gh plumbing only. That boundary is documented in the workflow itself.

Validation: 999 tests passing (+46 new), build/typecheck/zizmor clean, live `sync-harness --check` green after pinning, live smoke pass. Two deliberate mistakes tested: dist-tag reversion caught by `--check`; empty-version banner guard (was matching everything).

One setup note: automated PRs need a `SYNC_TOKEN` secret (fine-grained PAT, Contents+PRs write — same pattern as `VERSION_TOKEN`, since GITHUB_TOKEN-created PRs don't run CI). Without it, the daily job explains what moved and how to sync by hand instead of failing.

riesbri:
## Both fixes pushed (`09379a5`) — workflow-only, architecture untouched

**1. Peer check no longer gates the suite.**
The `released` job now *records* the peer verdict first (`ACCEPTED` / `REJECTED`, with checker crash mapped to `UNKNOWN` rather than impersonating either), then runs install → build → typecheck → full suite → consumer smoke independently (`continue-on-error` + recorded outcomes). One final `always()` step prints exactly:

```
runtime compatibility: PASS/FAIL
peer contract: ACCEPTED/REJECTED
```

…into the log and the job summary, and fails the job if **either** axis failed. So a newly published tuple rejected by the ranges — the expected state until a human approves it — still yields a full runtime answer, reading as "works, awaiting an explicit range decision". The rejection table is attached to the summary as a collapsible detail block. Ranges are still never widened automatically.

**2. Idempotent sync PRs.**
The PR job now maintains one rolling branch, `automation/harness-sync` (no more date-suffixed duplicates or same-day ref collisions): reset onto today's base, force-pushed (safe — the workflow concurrency group makes this job the branch's single writer), and its **open** PR is edited in place when it exists, created only when none does. Title/body still flip between "routine bump" and "peer compatibility decision required" per classification; the body notes the rolling nature so reviewers read the latest state.

Security boundary unchanged: executing jobs remain `contents: read` with no secrets; the write-capable job still executes no dependency code (manifest rewrites, `--lockfile-only` resolution, git/gh plumbing); peer ranges remain an explicit human decision.

Verified: zizmor clean on all workflows; verdict shell logic dry-run against all three decisive cases (all-green → green; REJECTED+green-runtime → red with legible distinction; FAIL+ACCEPTED → red).

PR #71, opened 2026-08-23
feat(plugins): /plugins — Harness preset composition browser

riesbri:
## Summary

Full `/plugins` implementation: a terminal control center over the running agent's Harness preset composition, matching Harness's own Web `/plugins` authority boundaries exactly.

- **`harness.ts`** — structural `AgentPresetsSeam`/`PluginsSettings` duck types via `ctx.get(...)`. Degrades cleanly when a profile mounts neither. Declares only the surfaces this frontend consumes: the real service's `remove(id)` is deliberately absent until a change actually offers deletion.
- **`composition.ts`** — recursive entry-list parser matching Harness's own (lenient) validator exactly: `id` is optional (not required — the Loader assigns a random one at mount when a row omits it), rows addressed by a structural `RowLocator` (index + name/id fingerprint per level) rather than id, since id is neither required nor unique. Tri-state `disabled` (`enabled`/`disabled`/`!!js`-conditional — conditionals are carried through for display and **never evaluated**). `toggleDisabled` re-verifies every locator step against a fresh re-parse before touching anything, refuses on a stale/reordered file, and returns the input verbatim (no re-serialization) when the requested state already holds.
- **`model.ts`** — `sessionBlank`/`resolveSessionPreset` (defensive-compatibility reimplementation, documented as such — not a blind trust of upstream's typed event), search, `toggleEligibility` (conditional refused before system/user trust — a `!!js` row on a built-in preset refuses directly, no pointless copy-then-refuse), `presetSwitchEligibility` (the `agent-preset-locked` boundary).
- **`catalog.ts`** — generation-stamped Harness reader mirroring `ConnectCatalog`. The roster's own `AgentPresetRow.broken` is authoritative over dshline's own (deliberately lenient) parser — a file this parser can make sense of is not proof of Harness health.
- **`actions.ts`** — `toggleRow` (the one narrow file adapter Harness exposes no seam for — gated on `trust === 'user'`, holds a real file lock across read/toggle/write, re-validates through `agentPresets.resolve()` afterward — Harness's own health check, not a second parse), `copyPreset`, `switchPreset` (gated on `sessionBlank`, logs `agent-preset/selected` with the id `recompose` actually returns), `setDefaultPreset` (`ctx.settings.mutate`, never a direct file write).
- **`overlay.ts`** — bounded composition browser. `/` is an explicit search-mode gesture (not dshline's usual always-on type-to-filter) — found live while implementing: `space`/`p`/`d` as bare shortcuts collide with search text the moment typing is unconditional, since composition rows routinely contain all three letters (`tool-pwsh`, `cordis`, `delegation`).
- **`index.ts`** (`openPlugins`) — orchestrates catalog + actions + overlay + the copy-confirmation/id prompts. A system-preset toggle offers a copy, suggests `<from>-custom`, and applies the same toggle to the fresh copy in one gesture. A blank session **positively confirmed** to be running the toggled/copied preset is recomposed live (and, for a genuine preset switch, durably logged); anything less certain — a started session, or dshline simply being unable to confirm what's current — is left as an explicit future-session customization, never a silent guess or a bypass of the lock.
- **attachment.ts** — registers `/plugins` per-agent, exactly like `/connect`/`/sessions`.

Invariant preserved throughout: Harness owns capabilities/composition/health/persistence/lifecycle; dshline only presents and invokes its contracts. No duplicate plugin registry, no provider-specific branches, system presets never modified in place.

### Two boundaries the review rounds sharpened

**Every eligibility fact is read at the instant it is acted on.** An action here holds its own awaits — two prompts a human answers, a file write, a Harness re-resolve — and a turn beginning across them has to move the answer. `liveEffectNote` and the copy-then-toggle path both read the session live (`sessionFactsOf`/`runningPresetId`) rather than from the catalog pass a keystroke was decided against. `liveEffectNote` is the module's only `recompose` that does not go through `switchPreset`, so it is the only one where a stale read could actually cross the started-session lock.

**A pre-preset session's composition is approximated honestly, and never claimed before it holds.** A session produced before this bundle adopted presets recorded no preset and resumes under `standard` — the preset built to mean the exact flat `dsh-base` set those sessions ran under. A deployment shipping no *usable* `standard` (missing, or present but broken — `resolve()` succeeds for a broken preset, so presence is not usability) falls back to its own default and reports the substitution, rather than refusing to open its own history. The caveat is emitted only **after** `mount` succeeds: `mount` rejecting rolls the resume back, and a line announcing "resumed under X" in the transcript of a session that never ran under X is a worse lie than the silence it replaced.

## Test plan

- [x] `npx tsc -b --force` — clean
- [x] `npx vitest run` — **1212/1213 passing** (one pre-existing skip: the opt-in Codex e2e)
- [x] `pnpm security` (`pnpm audit --audit-level moderate` + workflow lint) — clean
- [x] Every behavioral claim checked by breaking the fix on purpose and confirming the named test fails — including two test bugs found that way, one of which surfaced a fourth stale-read site that reading alone had missed
- [x] Packed fresh `pnpm install` of both published packages plus every declared peer: resolves with zero peer warnings, and a plain Node import of the built `@dshline/dshline` succeeds — confirms `yaml`'s and `dsh-atomic-write`'s peer metadata cause no install/runtime regression for an external consumer
- [x] Manual walkthrough against the real `deepseek-harness` runtime (`dsh --profile dshline`), with **no** manual profile patches: a brand-new session composed live from the roster's default (proving `mount()` ran inside `setup()`), then `/plugins` → search `codex` → Space → confirm copy → accept suggested `standard-custom` → row toggled → `p` → switch the blank session live → durable selection persists across reopen → `d` → set default → a genuinely new session composes from `standard-custom`. The shipped `standard` preset file was never touched. `--dump-config` confirmed `disabled: true` on every listed id, `agent-presets` present with `default: standard`, and `tool-ask-user` no longer double-inserted.

### Coverage of the risky paths

- **Composition editing** — `plugins-composition.spec.ts` includes a regression fixture drawn from the real `standard` preset: one toggle preserves both `!!js` rows, the persona block scalar, comments, and an untouched sibling group member. `lineWidth: 0` keeps the default 80-column fold from reflowing the whole file.
- **Untrusted text** — a preset's display `name` is file content (`preset.yml`), so it is escaped at the one choke point all three header call sites share; a name carrying `ESC[2J` is displayed, not obeyed, and no longer mis-measured by `displayWidth`.
- **Self-bounding frame** — the header block is two rows normally and three while the session runs a different preset than the one browsed (the state copy-to-customize produces), so its height is measured per render rather than assumed. A budget that assumed the shorter form built an over-tall frame that the height guard then collapsed to a one-line fallback, losing the whole browser at 90×11.
- **No path can crash the session** — a failure that isn't an answer (`Session.append` refusing an event) is caught rather than left as an unhandled rejection, and the recovery report is itself best-effort: drawing is this domain's only channel, so a throw from `report`/`commit` is swallowed rather than allowed to reject the promise the catch exists to settle.


riesbri:
## Update: dshline's agent plane is now presets-native by default

The manual walkthrough exposed that `/plugins` only worked because the local dev profile had been hand-patched to mount `dsh-agent-presets`. Fixed the actual shipped architecture, not just the feature on top of it — researched Web's own `packages/bundle/web-app/cordis.patch.yml` first and copied its recipe rather than inventing one:

- `cordis.patch.yml` disables the ~23 `dsh-base` rows that are per-agent tool rows a preset also lists (tool-bash, tool-fs, the tool-subagent family, tool-workflow, plan-mode, etc — same host-plane-vs-agent-plane reasoning in the comments, copied from upstream), then inserts `agent-presets` with `default: standard`. Registries/sandbox/token-meter/subagent-registry stay host-plane, untouched. dshline's own old `tool-ask-user` insert is removed — every shipped preset already mounts it.
- `window.ts`'s `attachOptions()` now composes every agent from its resolved preset inside the one supported `setup(agentCtx)` hook — a resumed session's own log wins over today's default (via the same `resolveSessionPreset` `/plugins` already uses), a fresh session's header gets stamped with the roster's default. No `agentPresets` seam mounted → no-op, flat composition exactly as before.
- `@deepseek-ai/dsh-agent-presets` joins as peer+dev dependency (same tier `dsh-tool-ask-user` already was) — confirmed empirically the running `dsh` CLI ships it bundled already, same reason `tool-ask-user` never needed a real dependency entry.

12 new tests (`window.spec.ts`, `bundle-patch.spec.ts` — a structural regression test against the actual shipped patch file, plus 2 new `sessions-plan.spec.ts` cases). 934/934 passing.

Verified beyond the suite: a packed fresh consumer install resolves clean with zero peer warnings; re-walked the real flow against the actual `dsh` CLI with **no manual profile edits** — a brand-new session composed live from the roster's default, `--dump-config` confirms every listed row `disabled: true` and no double-insert.

One environment quirk worth recording for whoever else runs this locally: a **stale globally npm-installed `@dshline/dshline`** shadows a profile's `link:`-declared local dev copy specifically for `cordis.patch.yml` reading at compose time (not for the plugin's own runtime JS, which correctly resolves the linked copy) — `npm install -g` after publishing this will pick it up automatically; nothing to fix in this repo.

riesbri:
## Update: legacy-session compatibility + two comment corrections

1. **Migration policy for pre-preset sessions.** Every session created before this PR has no recorded preset at all. \`mountAgentPreset()\` now distinguishes: recorded preset → always wins; no recording but the session already produced a turn → legacy, mounts \`standard\` specifically (not today's default, which may have moved to \`minimal\`/\`code\`/a custom preset since); no recording and still blank → today's default, same as any other new session. A non-stock deployment with no \`standard\` preset fails that resume outright rather than guessing — matches \`setup\`'s own documented rollback contract.
2. Fixed the \`tool-ask-user\` comment — it doesn't claim every preset mounts it anymore (\`minimal\` deliberately doesn't). No functional change.
3. Tightened the no-\`agentPresets\`-seam fallback wording in three places (\`window.ts\`, \`index.ts\`, the changeset) — the no-op only restores the old flat tool set for a deployment that never applied dshline's own disable list; removing just the seam from an otherwise-stock install leaves an agent with no tools at all.

4 new regression tests. 938/938 passing, \`tsc -b\` clean, packed fresh-consumer install re-verified clean.

riesbri:
Docs pass done. Reviewed README, usage, architecture, comparison, roadmap, install against the preset architecture.

- **README** — `/plugins` added to command list + capability paragraph (`ctx.agentPresets`).
- **docs/usage.md** — commands table row + new "Plugins" subsection: search, toggle, copy-to-customize, preset picker, default-setting, session-lock note.
- **docs/architecture.md** — capability-surfaces table row + new "Presets: composition is Harness's, not dshline's" section (system presets read-only/copy-to-customize, new/resumed/blank-session composition facts, agent-plane-behind-presets rationale mirroring Web).
- **docs/comparison.md** — "Work and session projections" → "Work, session projections, and agent presets."
- **ROADMAP.md** — new "5. Agent presets — merged" section (renumbered 6-9 after); two new Current-limitations bullets (session-lock, direct-file-edit mechanism); one new non-goal (no plugin marketplace, `dsh plugin` already covers that).
- **docs/install.md** — reviewed, no change needed (pure install mechanics, no runtime-architecture claims).

No docs-link-check script exists in this repo; manually verified every link/anchor touched or adjacent still resolves. Full suite (1196 passed/1 skipped), `tsc -b`, and consumer smoke (skipped on darwin, no `script(1)`; already verified via manual pack in the prior round) all clean.

PR #79, opened 2026-08-24
feat(dshline): keep turn timing live

riesbri:
## What changed

`/timing` no longer dumps a static bar chart under each finished reply. It now renders a **persistent, bounded live panel** in the live region:

- **On:** persistent whenever the live-region budget permits. During a turn it updates in real time — reasoning/output spans grow as chunks arrive, tools appear when called and freeze at their measured total on result, the header's wall clock ticks live (`timing · turn 3 · 41.2s · live`). While idle it retains the most recent finished turn; after a resume it shows `timing · no turn measured yet` rather than fabricating history from a replay that carries no `assistant/chunk` events.
- **Off:** every row disappears — nothing is shown at all.
- **Degradation:** timing is optional instrumentation and yields before critical interaction chrome. On an extremely short terminal its span rows go first, then its header; input and status stay usable to the end, and the composer sheds only its decorative separator before touching rows reserved for others.
- The finished chart is deliberately **no longer committed to scrollback**: with a persistent panel, keeping both read as two sources of truth for one measurement.

Visual style follows the Work overlay rows (dim label column, cyan bar, right-aligned duration), capped at six rows: header + four spans + `… +N more`, so the bounded live region cannot grow with tool count.

## Design decisions

Implemented by a Codex agent against five reviewer-proposed starting points; verdicts recorded here:

- **(a) Presentation — agreed, plain indented panel, no box**, as its own `timing` slot between `completion` and `status`.
- **(b) Data — agreed**: new `TurnTimer.snapshot()` yields incremental state; the existing attachment-scoped spinner heartbeat drives redraws (no second timer). Observation runs regardless of the preference, so enabling mid-turn is immediately truthful instead of inventing a partial turn from the toggle.
- **(c) End-of-turn — agreed to stop committing to scrollback**; panel retains the latest completed turn while idle.
- **(d) Bounds — agreed**: hard cap of six rows with elision; composer/completion reserve the panel's minimum row.
- **(e) Real-time vs log-timestamp principle — agreed with an explicit exception**, documented in the module header: completed measurements stay event-timestamp-derived; only open turns/tools tick provisionally against the wall clock because their closing event has not arrived.

One assumption needed refinement during implementation: `TuiSlots` subtracts rows spent *above* a view but reserved nothing *below* it, so merely adding and capping a slot would not guarantee persistent presence beside a tall composer or completion list. Those views now carry an explicit conditional downstream reservation. A review pass (`3e4fda8`) then closed the two gaps that reservation left: `cursor()` now receives the same unspent-row budget as `render()` (one shared calculation inside the composer), and the frame sheds decoration before structure on terminals too short for everything.

## Verification

- `pnpm build`, `pnpm typecheck`, full suite: **1443 passed / 1 skipped**; `pnpm audit` clean; zizmor (pedantic): no findings.
- xterm frame coverage (`packages/dshline/tests/timing-frames.spec.ts`): visible when on / absent when off, real-time growth between event batches, row capping, narrow terminals, resume placeholder, clean scrollback — plus, after the review commit: shared cursor/render budget with the hardware cursor asserted **on the visible tail cell** of a scrolled composer on a short terminal, empty/filled separator shedding with the cursor following, input surviving an impossibly small terminal while the panel yields whole, and composer + completion + timing + status held together on a fourteen-row terminal.
- Break-your-fix checks: inverted enable guard → named on/off frame test failed; cap raised 6→60 → capping test failed; snapshots frozen at turn start → growth test failed; restored fixed-10-row cursor → visible-tail regression failed by name; reverted separator shedding → named ladder tests failed by name.
- Minor changeset included (`red-cups-move.md`).

🤖 Co-authored implementation: Codex agent (design verdicts above); review fixes, verification, and PR upkeep by ox-alpha.

### DIFF
--- a/.changeset/red-cups-move.md
+++ b/.changeset/red-cups-move.md
@@ -1,0 +1,5 @@
+---
+"@dshline/dshline": minor
+---
+
+Replace post-turn timing dumps with a bounded live panel that tracks active turns and tools in real time.
--- a/docs/design.md
+++ b/docs/design.md
@@ -16,7 +16,7 @@ How this interface is built, and the reason behind each decision. Every heading 
 - [The status line drops details instead of cutting them](#the-status-line-drops-details-instead-of-cutting-them)
 - [Prices are shipped, and every one of them is wrong eventually](#prices-are-shipped-and-every-one-of-them-is-wrong-eventually)
 - [A command that takes a value offers its values](#a-command-that-takes-a-value-offers-its-values)
-- [The profiler does not claim the parts add up](#the-profiler-does-not-claim-the-parts-add-up)
+- [The profiler stays live without inventing history](#the-profiler-stays-live-without-inventing-history)
 - [Character widths follow the Unicode standard](#character-widths-follow-the-unicode-standard)
 - [Measuring and cutting agree about escape sequences](#measuring-and-cutting-agree-about-escape-sequences)
 - [Untrusted text is made safe before it is drawn](#untrusted-text-is-made-safe-before-it-is-drawn)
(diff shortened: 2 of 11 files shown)


riesbri:
## Review fixes pushed (`3e4fda8`)

**1. Cursor/render budget mismatch — fixed at the interface level.**
`TuiSlotView.cursor?(columns, rows?)` now receives the same unspent-row figure `render()` got; `TuiSlots.compose()` computes it once per view and hands it to both. Inside the composer a single `contentBudget` closure feeds both paths, so the drawn slice and the cursor's assumed slice cannot drift.

*Regression test:* short terminal (10 rows) + 15-line composer + timing on → asserts the hardware cursor sits **on the visible tail row** (`emulator.cursor()` + `emulator.screen()`/`cell()`), not merely that the frame looks right.
*Break-your-fix:* restoring the old fixed-10-row cursor calculation → `puts the cursor on the visible tail…` fails by name. ✅

**2. Empty-composer minimum rows — now intentional and shared.**
Both frame paths follow one ladder: decoration before structure. The separating blank sheds first when the reservation can't fit the full frame; borders never shed — below ~5 rows usability outranks the reservation, mirroring how the panel yields body rows before its header. The empty-frame cursor moves up with the shed separator via the same calculation.

*Tests:* separator shed at 5 rows / restored at 6 (ladder pinned), cursor row 2→1 contract, impossibly-small 4-row terminal keeps input + status while the panel yields, filled frame sheds too (30-line paste in a 5-row terminal stays on screen with honest `+29 rows`).
*Break-your-fix:* reverting either path to always-keep-separator → the three named separator tests fail. ✅

**Additional re-check:** new composed test mounts composer + real completion list + live panel + status together on a 14-row terminal (`/tim` mid-word), asserting screen height, header presence, and cursor placement. Also re-verified: enabling/disabling timing changes the reservation live (existing tests), narrow terminal, resume placeholder, scrollback cleanliness — no other gaps found.

**Results:** build ✅ · typecheck ✅ · full suite **1443 passed / 1 skipped** · `pnpm audit` clean · zizmor pedantic: no findings. Docs updated with one honest paragraph on the degradation order; changeset unchanged (still `minor`, scope unchanged). No architectural expansion — no global layout budget introduced.

PR #87, opened 2026-08-24
feat(dshline): let ctrl-o reach older truncated cards

riesbri:
Closes the `ctrl-o` limitation in `ROADMAP.md`.

Stacked on #85 → #84 → #83. It touches `docs/usage.md` and `ROADMAP.md`, which become bilingual pairs in #84, so it stacks rather than desyncing them.

## The problem

A compact card commits its elided rows straight into native scrollback, where `compact → full → hidden` cannot recover them — that cycle only affects cards drawn from then on. The inspector was their only way back, and it held **exactly one card**. The next tool call took the offer over, and a result you scrolled past was gone for good.

## The change

The last **twelve** truncated cards are retained, newest first.

- **The first `ctrl-o` is unchanged** — newest unseen card, one-shot. That is what keeps the detail toggle a single keystroke away, and it is why this is not simply "`ctrl-o` cycles through everything."
- **Reaching an older card is a second gesture**, `ctrl-o` from *inside* the inspector. It has to be handled there: a mounted overlay owns every key (`window.ts:376` routes to `activeOverlay.handleKey`), so the runner's own handler is not reached again until it closes.
- **The title counts the position** (`Tool output 2/6`). Two calls to the same tool can present almost identically; an overlay that changes with no sign of having moved reads as a redraw, not as navigation.
- **The hint appears only while an older card exists.** A hint for a key that does nothing reads as a failure rather than as the end of the history.
- **Stepping stops at the oldest** rather than wrapping, and drops the width cache and scroll offset with it — otherwise an older card opens partway down a body nobody has seen the top of.

## Invariants held

- The inspector is still a live-region overlay; **committed scrollback is never rewritten**. Asserted against a real xterm emulator, including across a step.
- `render(columns)` stays pure per card — the cache is dropped on step, with a test for it.
- **Bounded on purpose.** An unbounded list of retained call arguments and results is a second transcript, which `docs/architecture.md` refuses. Older than twelve, the elision marker beside the committed rows is the honest answer.

## A behaviour change worth flagging

A newer short result or an error **no longer discards** the retained history. That discarding existed only to stop one stale offer from capturing `ctrl-o` forever — which marking an offer consumed now handles instead.

Three tests asserted the old clearing contract. They are **rewritten, not deleted**, so the new contract is stated where the old one was:

| was | now |
| --- | --- |
| `clears the pending opportunity when a newer short result lands` | `keeps a truncated card reachable after a newer short result lands` |
| `clears the pending opportunity when a newer error result lands` | `keeps a truncated card reachable after a newer error result lands` |
| `leaves none pending after a replay where a newer non-truncated result followed` | `retains every truncated result of a replay, newest first` |

Plus new coverage for stepping, the stop-at-oldest boundary, the cap, and that walking the history does not leave a queue `ctrl-o` would re-offer one at a time.

## Verification

```
pnpm run build && pnpm run typecheck   # clean
pnpm run test                          # 1515 passed | 1 skipped (+45)
pnpm run verify-docs                   # 9 bilingual pairs consistent
```

Changeset included (`minor`, `@dshline/dshline`). Both `ROADMAP` and `docs/usage` counterparts updated and re-recorded — the gate went red first, as intended.

Not yet verified by hand against a live Harness profile; that is worth doing before merge.


riesbri:
Superseded by #89. GitHub auto-closed this PR when its base branch `docs/verified-platforms` was deleted on merging #85, and it could not be reopened afterwards because the head branch had been force-pushed in the meantime. #89 is the same branch and the same change, rebased onto `main`.

PR #88, opened 2026-08-24
feat(plugins): name the external server a row connects to

riesbri:
Stacked on #87 → #85 → #84 → #83.

## The MCP finding, which is not what I expected

Upstream ships `@deepseek-ai/dsh-mcp-client@0.1.1-rc.2` (`deepseek-harness/packages/mcp/mcp-client`). Reading it changes what is worth building:

- It is **one plugin instance per server**, `inject = ['tools']`, registering tools as `mcp__<serverName>__<rawName>`.
- It publishes **no ctx service** — no registry, no server list, no connection state.

Two consequences:

1. **MCP tools already render correctly today, with zero work.** Cards resolve definitions via `ctx.tools.get(name, agent)` and render `presentCall`/`presentResult` without knowing tool names. There was never anything to build there.
2. **A real `/mcp` browser cannot be built without inferring servers from the `mcp__<server>__` prefix** — inferring capability state from rendered identifiers, an explicit non-goal (`ROADMAP.md`, and "never inferred from tool names or rendered output" in the presets section).

So MCP is upstream-blocked in the same way roadmap items 6–9 are. **This PR does not build `/mcp`.**

## What it does instead

A row that connects to an external server takes its identity from *which server that is*. Two `dsh-mcp-client` rows load the same module and were indistinguishable in `/plugins`, because `summarizeConfig` shows nothing for a config with more than three keys or nested values — and an MCP config has a command, an argument list, and an environment.

`config.serverName` and `config.transport` are now read **structurally**, following exactly the pattern `config.provider` already set (`readConfigProvider`):

- named after the **fields**, not after any meaning
- read for **any** row that declares them, not for a module the parser recognises
- **never** read when the value is a `!!js` expression, which this module does not evaluate
- never computed for a group row

```
❯ ○   mcp-docs      @deepseek-ai/dsh-mcp-client      docs · stdio
  ○   mcp-search    @deepseek-ai/dsh-mcp-client      search · streamable-http
```

It reports what the composition file **says** — not that the server exists, is reachable, or is connected. Those are the questions no seam answers.

## Also

A `ROADMAP.md` limitation (both language sides) stating plainly that MCP servers have no browser because `dsh-mcp-client` publishes no registry, and that an upstream registry would make one possible. I will open that upstream issue separately — it is the actual unblocking action.

## Verification

```
pnpm run build && pnpm run typecheck   # clean
pnpm run test                          # 1521 passed | 1 skipped (+6)
pnpm run verify-docs                   # 9 bilingual pairs consistent
```

Changeset included (`minor`). Not yet exercised against a live Harness profile with two real MCP servers; worth doing before merge.


riesbri:
Now stacked on #89 rather than #87 — #87 was auto-closed when its base branch was deleted on merging #85, and #89 replaces it. The head branch this stacks on (`feat/ctrl-o-older-cards`) is unchanged, so this PR's diff is unaffected.

PR #91, opened 2026-08-25
feat(renderer): choose colour by role, not by name

riesbri:
Every coloured byte in the product went through `style(text, 'red')`, which names an appearance rather than a meaning. That is written identically for a failed tool (`transcript.ts`) and for a removed line of a diff (`cards.ts`), so no second palette could ever move one without moving the other — and four unrelated meanings (a warning, the spinner, context pressure, and every overlay border) shared yellow the same way.

That, rather than the absence of a picker, is what `ROADMAP.md`'s **"No themes. One color palette is currently shipped."** actually recorded: there was no way to *have* a second palette.

This is the enabling refactor. **No visual change**; `NO_COLOR` support is the only user-visible effect.

## The rule that makes ~256 call sites reviewable

The default palette emits **bytes identical to today's output**, including SGR parameter *order* — `markdown.spec.ts:312` matches `CSI 1;36 m` and `stream.spec.ts:178` matches `CSI 2;3 m` literally, so a role table written colour-first would render the same and fail the suite. Each entry is transcribed from its call site's argument order, never re-derived.

Under that rule **no existing colour assertion needed editing**. `theme.spec.ts` locks the property role by role, so any real colour change now lands separately, against a red test.

Verified beyond the suite: the same input rendered under `main` and under this branch, diffed as raw bytes — identical across every row.

## Design notes

**Why a module-level palette, not a threaded painter.** `HEADING_STYLES` and `EMPHASIS` are module-level tables inside `renderMarkdown`/`renderInline` — published pure functions that specs call directly. Threading a painter means a breaking signature change plus a painter parameter through ~25 view constructors and `slots.ts`, which has no such concept; it could not be reviewed as "no visual change". The palette is process-global for the reason raw mode is, and follows `acquireTerminal`'s rule: installing returns one disposer, safe to call twice.

**Why not just remap the `Style` table at startup.** It makes `Style.reset` remappable, and `width.ts` recognizes a closer with `RESET_PATTERN` and nothing else — a palette touching it would silently disable reopen-on-wrap. More fundamentally it still cannot express a role, only a colour.

**`gray` (113 uses) and `dim` (43) stay two roles** — `muted` and `subdued`. They are byte- and cell-different today, so merging is a visual change. Semantically `dim` composes with any foreground while `gray` is absolute and illegible on a light background. Merging is a one-way door.

## Widening past 16 colours costs nothing

`sgr()` is extracted so `style()` and `paint()` share one emitter, and `[38, 2, r, g, b]` joins the same way. **`width.ts` needs no change** — verified against all four of its consumers, not assumed: `ANSI_PATTERN` already swallows a multi-parameter sequence as one zero-width token, and its only real SGR *parsing* is reset detection. The one invariant is that the closer stays `CSI 0 m`; `CSI 39 m` renders identically and would be read as an *opener* by every wrap.

## Also included

- `NO_COLOR`, `FORCE_COLOR`, `COLORTERM`, `TERM=dumb` — none was read anywhere before. Resolved by a pure function in the renderer, with `process.env` read in `window.ts` where `acquireTerminal` already couples to real streams.
- A check failing the build if `style(` regrows a call site outside its own definition.
- `docs/design.md` gains the decision; `AGENTS.md` gains rule 3, beside the two colour rules it belongs with. Chinese counterparts patched and re-recorded.

## Verification

Build, typecheck, full suite (1538 passing), `verify-docs` (9 pairs consistent), `pnpm security`.

Broken on purpose, per `AGENTS.md`:
- `heading-1` flipped to `[36, 1]` → fails the golden lock by name
- closer changed to `CSI 39 m` → fails 8 tests across `theme`, `width`, `transcript`
- one `style()` reintroduced in `resume.ts` → the lockdown check fails and names `resume.ts:68`

**Not verified:** no real-terminal run. The pseudo-terminal scripts `AGENTS.md` describes are not in this repo, so live behaviour rests on the xterm-headless frames specs.


riesbri:
## Reworked after review: role ownership moved to the correct layer

`assistant`, `reasoning`, `tool-icon`, `mode`, `pressure-*` are what *this frontend* means by a colour. Declaring them in the renderer made a package that must not know what an agent is enumerate the parts of an agent session — the dependency-free rule broken from the other direction.

**The renderer now declares only the roles it draws itself**: markdown structure and generic emphasis, because rendering markdown is a renderer capability. `PaletteRoles` is an interface rather than a union, so `dshline` adds its 32 from its own package by augmentation. Still one `paint`, one `Palette`, one totality check — a role added without a colour still fails the build — but the vocabulary lives where its meaning does.

Two consequences:
- The renderer no longer publishes a complete palette (a consumer's augmentation makes `Palette` total over roles it knows nothing about). It publishes `MARKDOWN_ROLES` to spread; the shipped palette is assembled in `dshline/src/theme.ts`.
- `paint` draws an unknown role unstyled rather than throwing — which is what lets the renderer draw before any frontend wires it up.

## Colour depth is Node's answer now

Deleted our `resolveColorDepth`. `getColorDepth` already honours `NO_COLOR`, `FORCE_COLOR`, `COLORTERM`, and `TERM`, plus the CI variables and Windows build numbers a hand-written table forgets. Ours agreed with it on every case it covered — which is the argument for deleting ours, not for keeping it: the cases it *didn't* cover are the ones nobody would notice being wrong. One mapping remains, Node's monochrome `1` → the depth at which nothing is emitted.

## Vitest now runs three projects

Not tidiness. The renderer's specs must keep passing with **nothing installed**, because its independence is what lets every rule about widths and cutting be tested at all; a shared setup file installing this frontend's palette would quietly hide a renderer that had come to depend on it. `packages/dshline` owns its setup; `tests/` and `tools/` get a third project.

Checked file-by-file against the old collection — a first attempt at the split had silently dropped the six `tools/*.spec.mjs` specs. Restored.

The golden lock splits the same way: renderer locks its 14 roles, dshline locks its 32.

Verified: build, typecheck, full suite (79 files, 1538 tests).

PR #92, opened 2026-08-25
feat(dshline)!: add /theme with five shipped palettes

riesbri:
> **Stacked on #91.** Base is `feat/semantic-color-roles`; review that first. The diff shown here is this change alone.

Retires `ROADMAP.md`'s **"No themes. One color palette is currently shipped."**

`/themes` is named-or-asked, the idiom `/usage` and `/reasoning` follow: both paths meet at one resolve so a typed word and a chosen row cannot drift, and a name matching nothing reports what was on offer instead of opening a picker the reader did not ask for. It reuses the shared list overlay — no bespoke overlay, and no `preview` hook added to `SelectSpec`, which seven overlays depend on.

## The palettes

| | |
| --- | --- |
| `default` | Unchanged, byte for byte |
| `high-contrast` | Bright sixteen-colour, avoids `dim` and bright black entirely |
| `ember` | Warm, dark terminal (24-bit) |
| `tide` | Cool, dark terminal (24-bit) |
| `paper` | Light terminal (24-bit) |

The three deep palettes each declare their own sixteen-colour form **per role**, so a terminal that cannot show one gets a reviewed decision rather than a nearest-colour approximation — and the command names the fallback it used instead of degrading silently.

## A theme is nine colours, not forty-six

Writing one role at a time would make every palette a place for two roles to drift apart by accident — exactly how four unrelated meanings ended up sharing yellow. A theme declares a swatch; one mapping assigns all 46 roles from it. Adding a role updates every shipped palette at once and cannot leave one with a hole. `default` is exempt: it comes from the renderer unchanged, because its per-role argument order is transcribed from the old call sites rather than chosen.

## `paper` is the palette that justifies the `muted`/`subdued` split

On a white background bright black is nearly invisible while `dim` is merely quiet, so a light palette has to move the absolute grey and leave the attribute alone. One role could never have done both — there is a test that fails if they are merged.

`paper` also gives `magenta` its first use in the product. It was a dead entry in the style table for the whole life of the project.

## Showing the effect needed solving, not assuming

Committed rows are never rewritten, so a picker cannot preview a palette by repainting the transcript — and the shared list draws every row in one colour. Both usual ways to demonstrate a theme are unavailable.

So applying one **commits a short sample** — a prompt, a reply, a tool card with a diff, a failure, and the status line — directly beneath the differently coloured history. The comparison becomes the point instead of the confusion, and it costs no overlay.

The report states the scrollback caveat only when the palette actually changed, and stays quiet when a reader re-picks what they already had.

## Verification

Build, typecheck, full suite (1568 passing), `verify-docs`.

The sample is rendered at 24, 40, 80, and 200 columns and asserted to fit — it goes into real scrollback, so a row too wide wraps permanently and cannot be redrawn.

**Not verified:** no real-terminal run; see #91 for why.

## Deliberately out of scope

No user-authored palettes. `Role` is the contract a theme is written against, and publishing it before real palettes have exercised it would freeze a vocabulary nothing has tested — the mistake the roadmap already warns about for `TuiSlots`.


riesbri:
## Two review findings applied

**The sample block is gone.** It committed a fabricated reply, tool call, diff, failure, and status line into the reader's real scrollback on *every* switch — output no session produced, permanent, and afterwards indistinguishable from output one did. It solved a real problem (a picker cannot preview a palette by repainting a transcript that is never rewritten), but inventing a transcript is not an acceptable answer to it.

What replaces it: the report line itself, drawn in the palette just installed and carrying two roles rather than one flat colour, so it *shows* the change instead of only describing it. The live region below — composer, status line, timing panel — redraws in the new palette anyway, which was always the larger part of the answer.

A switch now costs **exactly one row** of scrollback, and there is a test for that number rather than for "more than five".

**`/themes` → `/theme`.** Every plural command here browses a collection (`/plugins`, `/profiles`, `/sessions`, `/todos`); every command that sets one value is singular (`/model`, `/reasoning`, `/usage`, `/timing`). This sets one value.

Also rebased onto the reworked role boundary in #91: the shipped palette now comes from `dshline/src/theme.ts`, and `builtin.ts` composes the other four on top of it.

Verified: build, typecheck, full suite (81 files, 1567 tests), `verify-docs` across all nine pairs. The confirmation line was rendered for `ember`, `paper`, and an unchanged `default`, at both 24-bit and the sixteen-colour fallback.

PR #93, opened 2026-08-25
feat(dshline)!: persist the theme through Harness settings

riesbri:
> Stacked on #92 → #91. Rebased after #94 merged.

`/theme` registers a `dshline` settings namespace and writes the choice into its user layer. Harness owns the document, the layering, the schema, the validation, and the change feed; this frontend owns one namespace and one key.

> **Supersedes this PR's original design.** It previously edited `~/.dsh/cordis.patch.yml` directly, because `connect/harness.ts` claimed the settings packages could not be added to this workspace. #94 disproved that — it was stale dist-tags, not upstream breakage. The workaround is deleted rather than polished: no YAML AST editing, no file lock, no atomic write, no `resolveConfiguredTheme` fallback parsing.

## Optional-consumer semantics come from Harness

`installSettingsSection` is the canonical wiring for a consumer whose settings service is optional. While a service exists it registers the namespace with this row's composition config as the `base` layer; when none is mounted, the source stays on that entry. **A profile with no settings provider still runs on its composed theme** — only saving is unavailable, and the command says so. None of that is conditional in dshline.

Layering is Harness's: schema default → `dshline` row's `config.theme` → `settings.yaml`.

## Invalid values are refused, not parsed around

The schema is a union of the shipped ids, not a free string. An unknown id is rejected before anything is stored. An externally edited document naming one fails registration and leaves the window on its composed theme — verified, not assumed.

## Persistence is retryable — review finding, and a real bug

Storing used to be conditional on `changed`. A switch whose write failed left the terminal already showing the theme and the document still naming the old one, and the obvious way to retry — picking the same theme again — was **the one gesture that did nothing**. It is unconditional now.

Still never rolled back on failure, for the reason `selection.ts` gives: the palette has already changed by the time the write runs. Two tests cover it; both fail against the old conditional.

## One authority

`WindowPrefs.theme` is gone — it had become a second copy of a fact the settings section owns. `window.palette()` remains as presentation state, documented as agreeing with the choice except in that failed-write case.

## Dependencies

Follows how `dsh-agent-presets` consumes the same service: `@deepseek-ai/dsh-settings` as a **peer** on the house range `^0.1.0-rc.7 || ^0.1.1-rc.2`, `@deepseek-ai/schemastery` as a **dependency** because a schema is a value this plugin constructs. `installSettingsSection` verified present at both ends of that range.

## Tests

Against the **real settings service**, not a fake — every behaviour that matters here is Harness's. 13 cases: layering, absent provider, storing, rejected id, storage failure, live notification, watcher disposal, externally stored invalid value.

## Real-terminal smoke test

Driven through an actual pty against an isolated `DSH_HOME`:

| | |
| --- | --- |
| switching | `· theme: ember · rows above keep the colours they were printed with` |
| four consecutive switches | each reported; the repeat says `— already in use` |
| persistence | `settings.yaml` gains `dshline: {theme: ember}` |
| across restart | fresh process boots painting ember truecolor; the default's bold-cyan banner absent |
| external edit, live | writing `theme: paper` into `settings.yaml` mid-session repainted the window with no keystroke — 1327 bytes, the live region only, not a history repaint |
| `NO_COLOR=1` | **zero** colour or attribute sequences emitted |


riesbri:
## Persistence rearchitected — `cordis.patch.yml` is gone

#94 disproved the assumption this PR's persistence was designed around, so the workaround is **removed rather than polished**.

`/themes` now registers a `dshline` settings namespace and writes the choice into its user layer. Harness owns the document, layering, schema, validation, and change feed; this frontend owns one namespace and one key.

### What replaced what

| before | now |
| --- | --- |
| YAML document-tree edit of `cordis.patch.yml` | `ctx.settings.mutate(ns, [{op:'set', path:['theme'], …}])` |
| `withFileLock` + `writeFileAtomic` | Harness's serialized write queue |
| `resolveConfiguredTheme` fallback-and-warn parsing | schema union of the shipped ids |
| `"saved for the next window"` | applies **live**, this window included |
| `WindowPrefs.theme` | removed — it was a second copy of the choice |

Deleted: `themes/persist.ts`, its spec, the AST editing, the lock, the atomic write, and the caveat path in `index.ts`. `yaml` and `dsh-atomic-write` remain dependencies — `/plugins` and `/profiles` still use both.

### Optional-consumer semantics come from Harness, not from a conditional here

`installSettingsSection` is the canonical wiring for exactly this case: while a service exists it registers with this row's composition config as the `base` layer; when none is mounted it leaves the source on that entry. A profile with no settings provider still runs on its composed theme — only saving is unavailable, and the command says so.

Layering is Harness's: schema default → `dshline` row's `config.theme` → `settings.yaml`.

### Invalid values are refused, not parsed around

The schema is a union of the shipped ids, not a free string. An unknown id is rejected before anything is stored. An externally edited document naming one fails registration and leaves the window on its composed theme — verified, not assumed.

### Live

`settings.yaml` edited by hand while a session runs repaints the window through the change feed. Committed scrollback is untouched, as always. The watch is guarded on the id because it fires for our own write too.

### Dependencies

Follows how `dsh-agent-presets` consumes the same service: `@deepseek-ai/dsh-settings` as a **peer** on the house range `^0.1.0-rc.7 || ^0.1.1-rc.2`, `@deepseek-ai/schemastery` as a **dependency** because a schema is a value this plugin constructs. `installSettingsSection` verified present at both ends of that range.

### Tests

Against the **real settings service**, not a fake — every behaviour that matters here is Harness's, and a fake would only prove the file agrees with itself. 13 cases: layering, absent provider, storing, rejected id, storage failure, live notification, watcher disposal, and an externally stored value the schema refuses.

### One thing to know

`window.palette()` (presentation) and `themeSettings.current()` (the choice) agree except in one documented case: a switch whose write failed has already changed the terminal and is not put back. That is `selection.ts`'s contract, followed here for the same reason.

Also rebased onto `main` now that #94 has merged; the stack order is unchanged.

PR #113, opened 2026-08-28
feat(dshline): let /connect declare custom Harness routes (Connect 2.0)

riesbri:
## Summary

- `/connect` can now declare a custom Harness route — a self-hosted server, a private gateway, a local endpoint speaking a protocol the mounted adapter supports — through `llm-pi-ai`, the one namespace whose settings profile can describe a whole provider route today. A new `+ Add custom provider` entry walks through provider id, endpoint, protocol (read from the namespace's own schema), an API key (only when the schema names somewhere to store one), and model catalog; the flow ends in an explicit review (Create provider / Cancel) before anything is written. An existing declared route gains an `Edit endpoint and models` action.
- Model discovery goes through `ctx.llm.discoverModels()` only — no `fetch()`, no provider SDK, no network code in dshline. Results are candidates a reader adopts, never written automatically; an id already in the draft is never overwritten by a discovered one.
- Connect's live refresh now coalesces `settings/updated`, `settings/document-updated`, `credentials/reference-updated`, and `credentials/record-updated` alongside `llm/adapters-updated`, so an edit from the web Models page or a hand-edited `settings.yaml` no longer needs `ctrl-r`.
- Architecture: the generic Connect plane (`ctx.llm`/`ctx.settings`/`ctx.credentials`/`ctx.authorization`) stays fully namespace-agnostic. `connect/model.ts` never infers a declaration capability from schema shape alone — `ConnectNewRouteTarget` is a plain data shape it never produces itself. That one Harness-specific fact — that `llm-pi-ai` specifically is a domain whose profile can describe a whole route — is asserted exactly once, in `pi-ai.ts`'s `piAiDeclarationTarget()`, which is scoped to `llm-pi-ai`'s own directory entries and gated on the curated fields (`baseURL`, `models`) and protocol choices this module actually depends on still being reachable. `+ Add custom provider` is offered only when every check passes, never a row guaranteed to fail partway through the wizard.
- Fails closed on schema drift: if the protocol/model/baseURL fields the create flow depends on stop being reachable, or a credential-reference field disappears, the wizard adapts (skips the key step) or refuses outright — never a guessed `api: ''`, and never a typed key silently discarded after a successful-looking write.
- No new runtime dependency on `@deepseek-ai/dsh-llm-pi-ai`, no provider registry, no second secret store. Harness floor unchanged (`0.1.1-rc.2`) — `discoverModels` and all four events are present at the pinned generation.
- `ROADMAP.md` and `docs/architecture.md`, with their Chinese counterparts, updated to describe the new declaration path and its boundary.

## Review history

This PR went through two review passes before merge (see inline comments/commits `603f263`, `93a5b18`):
- Pass 1 fixed an inherited-vs-explicit-empty models bug, replaced a too-generic schema-shape inference with the namespace-scoped `piAiDeclarationTarget()`, added a fail-closed protocol check, a global (live + configurable) duplicate-id check, API-key normalization, revision-at-draft-open anchoring, and the explicit final review menu.
- Pass 2 hardened `piAiDeclarationTarget()` further (models-field reachability), fixed a silent-key-drop hazard (the wizard no longer asks for a key it has nowhere to store, instead of asking and discarding the answer), and made "Create" with zero models selected stay inside the review draft instead of closing the wizard.

## Test plan

- [x] `pnpm -w typecheck` — clean
- [x] `pnpm test` — 460/460 suites, 1944/1945 tests (1 pre-existing unrelated pending)
- [x] `node tools/check-peer-currency.mjs` — every peer `ok` against the currently published line
- [x] `node tools/verify-translation-pairing.mjs` — all 9 bilingual pairs consistent
- [x] `connect-model-editor.spec.ts`, `connect-pi-ai.spec.ts` (including `piAiDeclarationTarget`'s full gate: dict-shape, parent-agreement, baseURL/models reachability, protocol derivability, and that a missing credential-ref field does *not* disable declaration), `connect-route-editor.spec.ts` (prompt-driven integration tests: discovery-through-`ctx.llm` only, settings-before-credential ordering, no secret leakage, normalized-key propagation, revision-at-draft-open, global duplicate-id collision, fail-closed protocol, keyless-schema creation, zero-model review recovery, inherited-vs-explicit-empty models including the submenu no-op regression)
- [x] Extended `connect-catalog.spec.ts` (live-refresh events + teardown), `connect-overlay.spec.ts` (the create-route row), `connect-model.spec.ts` (`newRouteIdProblem`)
- [ ] Manual acceptance against a real local OpenAI-compatible endpoint — no such server was reachable in this sandboxed environment (checked common local model-server ports; none responded); the generic path is proven at the seam level by the fixtures above instead


riesbri:
Pushed a fix commit addressing every point.

## 1. Inherited-model bug (merge blocker) — fixed
`editModels()` now returns `{ entries, changed }`. `changed` is measured as "would this write something different from what's there now" (via the existing `sameModelSet` comparison), not "was the submenu entered." `runRouteEditor` only flips `modelsInherited` to `false` when `changed` is true. Regression coverage added: inherited→Models→Done→Save (no mutation, still inherited), inherited→fetch→adopt none→Save (no mutation), inherited→adopt one→Save (explicit array written, `reset-models` now offered).

## 2. `declarableTargets()` fake genericity — fixed
Moved the whole determination into `pi-ai.ts`'s new `piAiDeclarationTarget()`, scoped to `llm-pi-ai`'s own directory entries only, and gated on: entries agree on parent path, schema shapes it as a `dict`, the curated `baseURL` field is reachable, and a protocol choice is derivable. `connect/model.ts` no longer infers anything from schema shape — `ConnectNewRouteTarget` is a plain data shape now, never produced by generic code. `catalog.ts` and `index.ts` simplified accordingly (no more namespace-filtering in `index.ts`, since every target in `row.targets` is now pre-vetted).

Agreed with the framing — this was the right call, not a "prefer genericity" tradeoff.

## 3. Fail closed on protocol derivation — fixed
`runCreateRoute` now checks `protocolOptions.length === 0` right after reading the fresh descriptor (before any field prompts) and refuses immediately with a named-namespace message. No `api: ''` fallback. Same check is now also load-bearing in `piAiDeclarationTarget()`'s gate, so in practice the row shouldn't even appear — this is the defensive second layer for the race between "row shown" and "wizard opened."

## 4. Global duplicate-id check — fixed
`taken` is now `listProviders().map(id) ∪ listConfigurableProviders().map(provider)`. Added a test: a live route with no configurable-provider entry still blocks a same-id custom route before the settings write.

## 5. Normalized key — fixed
Was a real bug — raw `apiKeyTyped` was reaching both `discoverModels()` and `credentials.set()`. Restructured so the draft only ever holds the post-`normalizeApiKey()` value; nothing downstream sees the raw typed text. Added a whitespace-trimming test.

## 6. Revision anchored to draft-open — fixed
`runCreateRoute` now writes with `descriptor.revision` (read fresh right as the wizard opens, before any prompts) instead of the `target.revision` carried from the catalog's last read. Added a test with deliberately different stale/fresh revisions.

## 7. Explicit final confirmation — added
New review menu before the first write: Provider ID (fixed), display name / base URL / protocol / API key ("configured"/"not set", never the value) / model count, each editable, plus Create/Cancel. Leaving the models submenu returns here, not to a write. Chose to keep this structurally separate from `runRouteEditor`'s loop rather than force a shared abstraction — the two differ enough (id immutability, the create-vs-edit terminal action, discovery identity shape) that a shared implementation would've needed its own branching to stay clear. Happy to reconsider if you'd rather see it unified.

## 8. Protocol-neutral wording — fixed
"The OpenAI-compatible endpoint/base URL this route calls" → "The base URL this route calls" in both the create wizard and the route editor.

## 9. Changeset wording — fixed
"No provider-specific dshline code" → explicit acknowledgment that `connect/pi-ai.ts` is provider-family-specific presentation knowledge, narrowed to "no provider runtime, SDK, transport, HTTP client, or secret store."

## Additional bug found during the pass
None beyond what's listed above, but worth flagging: while normalizing the API key (#5), the discovery request for a *new* route was also passing the raw typed key — same root cause as the credential-store leak, same fix.

## Verification
- `pnpm -w typecheck` — clean
- Full workspace: 460/460 suites, 1941/1942 tests (`success: true`; 1 pre-existing unrelated pending)
- `node tools/check-peer-currency.mjs` — all `ok`, floor unchanged at `0.1.1-rc.2`
- `node tools/verify-translation-pairing.mjs` — all 9 pairs consistent
- New/changed tests: `connect-route-editor.spec.ts` (+7 tests: global-collision, fail-closed-protocol, revision-anchoring, review-cancel, review-confirmation-no-secret, 3× inherited-models regressions), `connect-pi-ai.spec.ts` (+8 tests for `piAiDeclarationTarget`'s gate), `connect-model.spec.ts` (removed the 4 tests for the deleted `declarableTargets`)

Changed files: `connect/{model,catalog,index,pi-ai,route-editor}.ts`, `docs/architecture.md`+`.zh.md`, the changeset, and the three test files above.

riesbri:
Pushed a final hardening commit (\`93a5b18\`) addressing all 5 points.

## 1. Schema-drift hardening — fixed, and a real bug found

**Models field**: added a check to `piAiDeclarationTarget()` — the curated `models` field must be reachable at a sample route's path, since `runCreateRoute` always writes one. Missing it now disables `+ Add custom provider` the same way a missing `baseURL` or protocol union already did.

**Credential-reference field — this was a real bug, not just a hardening gap.** Before this fix: the wizard always prompted for an API key regardless of whether the schema named anywhere to store one. If it didn't, the key was silently dropped after settings.mutate() succeeded, and the outcome still reported `route created` — no error, no indication the key went nowhere.

Chose your preferred design's *spirit* but the *conditional* variant, not the disable-outright one: absence of a credential-ref field is legitimate (unauthenticated local servers are a real, wanted use case per the original brief), so `piAiDeclarationTarget()` does **not** require it. Instead `runCreateRoute` reads whether the field exists once, before any prompt runs, and when it doesn't: skips the API-key prompt and its review row entirely. The question is never asked when there's nowhere to put the answer — no accept-then-drop path exists anymore.

Tests added: `piAiDeclarationTarget` gate tests for missing-models (disables) and missing-credential-ref (does *not* disable, confirms the design choice), plus a `runCreateRoute` integration test proving the key prompt/review-row are both absent and `credentialCalls` stays empty for a keyless-only schema.

## 2. Zero-model Create stays in review — fixed
Selecting "Create provider" with nothing selected now sets a notice (`Select at least one model before creating this provider.`) shown in the existing review menu's `detail`, and `continue`s the loop instead of returning `failed`. Added a regression test: zero models → Create → refused in place, no mutation → open Models → add one → Create again → succeeds. The old "at least one model" post-loop failure path is now unreachable and was removed.

## 3. PR description — rewritten
No longer mentions `declarableTargets()`. Now describes the actual final architecture (generic plane + `piAiDeclarationTarget()` as the one scoped exception), the two-pass review history, and current test counts.

## 4. Real endpoint acceptance — not run, honestly
Checked for a local OpenAI-compatible server on the common ports (11434, 1234, 8000, 8080) — nothing responded in this sandboxed environment. Did not fabricate a run, and did not add any code (Ollama/LM Studio detection, localhost scanning, etc.) to try to make one appear. The generic path stays proven at the seam level by the fixture-driven integration tests instead.

## 5. Final verification
- `pnpm -w typecheck` — clean
- `pnpm test` — 460/460 suites, 1944/1945 tests (`success: true`; 1 pre-existing unrelated pending)
- `node tools/check-peer-currency.mjs` — all `ok`, floor unchanged at `0.1.1-rc.2`
- `node tools/verify-translation-pairing.mjs` — all 9 pairs consistent
- CI, security, and CodeQL all green on the pushed commit

## Additional issue found this pass
The credential-ref silent-drop under #1 — flagging it explicitly since it was a genuine bug (not just missing hardening): a typed key could be discarded while creation reported success. Fixed as described above.

PR #130, opened 2026-09-01
feat(dshline): set the profile up on a first run, then continue into it

riesbri:
## What a user saw

```sh
npm install -g @deepseek-ai/dsh @dshline/dshline
dshline
# dshline: the "dshline" profile does not exist yet. Create it once with:
#
#   dshline --setup
```

The one-word command was two words on every new machine, and the refusal arrived after the user had already asked for a session.

## Now

With no `dshline` profile yet, `dshline` asks once (default yes), runs `dsh plugin --profile dshline add @dshline/dshline` through the launcher it just resolved, and continues into the launch that was typed — `dshline --resume`, `dshline -C ~/project`, `dshline "run the tests"` all land in the session they asked for, argv boundaries intact.

```
dshline: first run — the "dshline" harness profile is not set up yet.

This will run:

  dsh plugin --profile dshline add @dshline/dshline

The harness creates the profile and installs this package into it, which uses
the network through pnpm.

Set it up now? [Y/n]
```

## Harness keeps the lifecycle

`bin/dshline.mjs` writes no profile file, never calls pnpm, never reads a package's `dsh.bundle`, and reconciles nothing. That one command is the whole of its involvement — `dsh plugin` already initializes a profile on first use and reconciles the bundle list against what is installed.

Three boundaries the obvious version of this gets wrong:

- **Uninitialized = no profile `package.json`**, the same test `dsh plugin` applies. The old check asked whether the *directory* existed, so an interrupted install left a state the wrapper refused to set up and Harness considered uninitialized.
- **An existing profile is never repaired** — missing dependency, empty `node_modules`, unresolvable package, malformed bundle list, a profile that will not boot all launch anyway, and Harness's loader says what is wrong.
- **Explicit `--profile` opts out entirely**, `--profile dshline` included. Fixes the old behaviour where `dshline --profile other` still checked the dshline profile and refused to start.

Non-interactive runs stop with `dshline --setup` guidance rather than installing packages a script never agreed to. `--setup` still asks nothing, needs no terminal, and still forwards a source path.

## Permission is for a command, not for a conclusion

A manifest that appears while the question is on screen means a setup **started**, never that one finished: `dsh plugin` writes the profile manifest before it installs anything. So there is no short-circuit — once the answer is yes, the authorized command runs. Skipping on that file would launch the frontend into a profile still being installed, and telling the difference apart would mean reading dependencies, `node_modules`, or bundle state: profile health, which this wrapper leaves to Harness by design.

Two overlapping confirmed first runs therefore both delegate. That is Harness's own concurrent-mutation question; dshline's promise is narrower and testable — a failed setup fails that invocation and launches nothing, and there is no lock or lifecycle state machine here. (Two concurrent `dsh plugin` mutations on one fresh profile can lose one process to a torn read of the manifest the other is writing; measured against the local checkout and against the published `latest` and `alpha` launchers. Worth an upstream issue — atomic manifest writes and mutation locking belong beside the code that owns the file — not a second package manager here, and not a permanent fixture in the compatibility lanes.)

## Windows

An npm install on Windows provides the launcher as a `dsh.cmd` shim. Spawning the bare name `dsh` finds no file there, and `spawn` has refused to run a batch file directly since the CVE-2024-27980 hardening — so PATH lookup now answers with the file it found (wrapper and `src/launcher.ts` alike) and a `.cmd`/`.bat` goes through `cmd.exe`, escaped the way `cmd` requires: qntm's backslash-and-quote rule (qntm.org/cmd) plus cross-spawn's `^` layers, transcribed rather than imported since the wrapper imports nothing, each argument escaped **twice** because a shim re-invokes its target with `%*` and the line is parsed a second time, and the whole command line wrapped for `/s`. Never `shell: true`, which joins argv unquoted; nothing about Unix spawning changed.

A quoting layer is worth exactly as much as the argv that comes out the other side, so a focused **`windows-latest`** job runs the two launcher specs on Windows, where the fixture's `$DSH_BIN` *is* an npm-style `dsh.cmd` and the stub records the argv it received after the shim:

| Proven on windows-latest | |
| --- | --- |
| `--setup` reaches the shim | argv `plugin --profile dshline add @dshline/dshline` |
| ordinary launch reaches it, found on **PATH** | argv `--profile dshline --cwd <cwd>` |
| first run then launch, both through the shim | setup argv, then launch argv |
| one task with spaces | one argv entry |
| double quotes | survive |
| `&`, `\|`, `^`, `%…%`, `!`, `( )`, `;`, `,`, `<`, `>`, backtick, `*` | literal argument data |
| trailing backslash, trailing quote, doubled backslashes before a quote | uncorrupted |
| all of them in one task | one argv entry |
| a task containing a line break | refused, exit 1, nothing spawned |

That matrix is not decoration: it caught a real bug in the escaping. cross-spawn matches the backslash run before a quote with a lazy group inside a lookahead, which for two or more backslashes doubles only the last one — `a\\"b` reached the launcher as `a\\b`, quote dropped by `CommandLineToArgvW`. The code now follows qntm's rule (double every backslash before a quote, then escape the quote) and a unit case pins the shape so the regression costs one assertion rather than a Windows run.

A line break inside an argument is refused on that path instead of being passed through: a `cmd` command line has no representation for one — the character ends the command — and handing user text to `cmd` as syntax is the failure the quoting exists to prevent. Every other platform, and a real `.exe` on Windows, take the argument unchanged.

Two things that job also caught: `.gitattributes` now pins LF checkouts (Git for Windows converts to CRLF, and a CRLF `bin/dshline.mjs` neither survives the test transform nor keeps a runnable shebang on Unix), and the ctrl-c regression test signals the wrapper directly rather than through a pty, whose SIGINT reaches `script(1)`'s own shell and makes its death the run's exit status.

## Also fixed

- **`dshline --version` / `-V`** now answer with this package's version — no harness, profile, terminal, or network needed. Before, the flag reached the app's commander and printed `error: unknown option '--version'`, which is exactly what the bug-report template asks contributors to run.
- **Signal ownership is lifecycle-local**: the SIGINT ignore is installed for one child and removed by name when that child settles, so a signalled child's status is re-raised without `removeAllListeners` deleting listeners this file never added. The property it protects is tested: with the stub frontend holding the terminal, `ctrl-c` reaches the child and the wrapper survives to report its clean exit.
- Making the raw wrapper importable for focused tests needed a symlink-safe executable-entry guard. npm's global bin is a symlink and `import.meta.url` names the target, so the first version of that guard did nothing at all when run through one — caught during development, and now covered by a test that invokes the wrapper through a symlink.

## Evidence

Deliberate mistakes, each shown to fail by name:

| Deliberate mistake | Result |
| --- | --- |
| detect the directory instead of the manifest | 2 cases fail |
| let an explicit `--profile` bootstrap | 3 fail |
| repair a broken existing profile | 8 fail |
| drop the post-setup launch | 7 fail, plus the consumer e2e |
| treat an appearing manifest as a finished setup | the two new concurrency cases fail |

- `packages/dshline/tests/launcher-bootstrap.spec.ts` — real child processes, a stub launcher that copies `dsh plugin`'s **ordering** (manifest first, install after), isolated `DSH_HOME`, and a real pseudo-terminal for the question (self-skipping where there is none). Covers yes/Enter/no/ctrl-c, non-TTY, `--setup` shapes, setup exit codes and signals, argv preservation, `DSH_HOME`, both `--profile` bypasses, no-repair, `--version`, resolution precedence, symlink invocation, SIGINT ownership, and the concurrency cases above.
- `launcher-policy.spec.ts` keeps the wrapper/`src/launcher.ts` drift guard, now including the Windows-shim case.
- Two consumer smokes, one question each:
  - `tools/consumer-smoke.mjs` (unchanged in purpose): does the **plugin code in this commit** install and boot against this Harness line? Packed tarball, published launcher, renderer fallback for the window between a version bump and its publish.
  - `tools/consumer-smoke.mjs --bootstrap` (new): does **this commit's wrapper** implement the user lifecycle? Packed executable, published launcher on PATH, a genuinely empty `DSH_HOME`, real pty, question answered, Harness creating and installing the profile, banner, `ready`, clean ctrl-d. The banner is checked against the version Harness actually installed, read from the profile it built.

    The earlier version of this smoke seeded that profile — initialize through Harness, append pnpm overrides, delete the manifest — to make `@dshline/dshline` resolve to the local tarball. All of it is gone; the profile package here is whatever the registry serves, which is what a real first run gets.

### CI

| Lane | Result |
| --- | --- |
| Core · Node 22.19 / 24 / 26 | green |
| Docs · EN/ZH pairing | green |
| Windows launcher · npm `.cmd` shim (new) | green — 67 passed, 0 failed, 19 skipped (the pty cases and three POSIX-signal/symlink semantics, each skipped with its reason) |
| Harness · Minimum (0.1.1-rc.2, floor launcher pinned by version) | green — first run installed `@dshline/dshline@0.13.0`, which is the version that resolves against the floor, and booted |
| Harness · Released | green — packed-plugin boot **and** first run from an empty home, banner `0.14.1` |
| Harness · Alpha (`@deepseek-ai/dsh@0.1.2-alpha.3`) | green — same two |

Locally: `pnpm build`, `pnpm typecheck`, `pnpm test`, `pnpm security`, `pnpm run verify-docs`, both consumer smokes, and the focused launcher specs.

Two CodeQL alerts on the launcher hand-off (`js/indirect-command-line-injection`, `js/shell-command-injection-from-environment`) are dismissed as false positives; the spawn is explicit `shell: false` with an argv array, and injection-negative tests on both POSIX and real Windows shims assert shell-looking task text causes no side effect. Full rationale is in the PR comment.

## Upstream notes (no Harness changes here)

1. `dsh plugin` takes no profile-mutation lock and writes the profile manifest non-atomically, so two concurrent mutations can lose one process to a torn read (and two *different* packages added at once can lose one from `dependencies` while both exit 0).
2. `runPlugin` spawns pnpm with `shell: true` on Windows and unquoted arguments; a spec containing a space breaks.
3. Under `DSH_HARNESS`, `dshline --setup ./packages/dshline` anchors inside the harness checkout (pnpm links the missing folder with only a WARN). Documented rather than papered over — rewriting it would mean owning a copy of Harness's package-spec parser.


riesbri:
## Security rationale for the dismissed CodeQL alerts (#15, #16)

Two code-scanning alerts on this PR are dismissed as **false positive**. Full audit trail, kept here because the API caps an alert comment at 280 characters.

### The alerts

| Alert | Rule | Reported flows (from the PR's SARIF) |
| --- | --- | --- |
| #16 | `js/indirect-command-line-injection` | `process.env.PATH` (`onPath`) and `process.env.DSH_BIN` into the `cmd.exe` plan and the final `spawn`; the message also cites the forwarded `process.argv` |
| #15 | `js/shell-command-injection-from-environment` | `process.cwd()` — the wrapper's own `--cwd` argument — into argv assembly and the final `spawn` |

Both sink at the single `spawn` in `packages/dshline/bin/dshline.mjs`. Neither exists on `main`: they appear with this PR because the Windows `.cmd` branch is the first place a real shell interpreter and a constructed command line exist at all.

### Why this is the intended boundary, not a vulnerability

dshline is deliberately a launcher wrapper. `DSH_BIN`, `DSH_HARNESS` and `PATH` **are** the mechanisms by which a user selects the Harness launcher — `DSH_BIN=/path/to/foo dshline` means "run `/path/to/foo` as the launcher" — and `process.argv` is forwarded to that launcher on purpose. Anyone who controls this process's environment already chooses the program this process runs; sanitizing those values would remove supported workflows without removing any authority.

What the code guarantees at the boundary:

- **The final `spawn` never asks Node to interpret argv through a shell.** `shell: false` is now written out explicitly at the call site, and there is exactly one `spawn` in the file — no `exec`, `execSync`, `execFile`, `spawnSync`, and no `shell: true` anywhere in the package.
- **On every platform except a Windows batch shim**, `spawnPlan()` returns the executable plus an argv array unchanged; user task text stays individual argv entries.
- **The one shell-mediated case** is a Windows npm `.cmd`/`.bat` shim, which `spawn` has refused to run directly since the CVE-2024-27980 hardening. It becomes `%ComSpec%` (or `cmd.exe`) with `/d /s /c "<line>"` and `windowsVerbatimArguments`, where each argv entry is context-escaped for the parse layers that shim actually causes — twice, because the shim re-invokes its target with `%*`, and the launcher path once, because the path takes part in the first parse only.
- **CR/LF in an argument is refused** on that path rather than passed through: a `cmd` command line has no representation for a line break, so the character would end the command.
- **`ComSpec` is honored deliberately.** Node's own `shell: true` resolves `process.env.ComSpec || 'cmd.exe'`, as does `cross-spawn`. Under the same trust boundary that already lets the environment pick the launcher outright, honoring it grants no additional authority, while hardcoding `%SystemRoot%\System32\cmd.exe` would break legitimate alternate-root and 32-bit setups for no security gain.

### Evidence, by side effect rather than by inspection

- **POSIX** (`shell syntax in a task`): tasks that would run a command if anything interpreted them — `; touch <marker>`, `&& touch`, `$(touch …)`, backticks, `| tee`, `> marker`, and one containing a real newline — each asserted to reach the stub launcher as **one literal argv entry**, with the marker file asserted absent afterwards. The same for a `--setup` source spec, the other place a caller's text reaches the launcher. Markers live inside the test's temporary fixture, and nothing in the proof runs a shell.
- **Windows** (`windows-latest` job, through the real `dsh.cmd` fixture, where `cmd` parses the line twice): `normal & echo injected > <marker>`, `&&`, `|`, `^&`, and `& type nul > <marker>` — one literal argv entry each, no marker created. Alongside the existing matrix that records the argv the shim's target received for spaces, quotes, `& | ^ %…% ! ( ) ; , < >`, backtick, `*`, trailing backslashes, doubled backslashes before a quote, all of them at once, PATH discovery, `--setup`, and the CR/LF refusal — plus a shim whose own directory carries a space and `& ( ) , ^`.
- **Broken on purpose:** with `shell: true` at the spawn, both POSIX cases fail, the task arriving as several argv entries instead of one. The tests are real.

### Why dismissal rather than further change

CodeQL cannot model the custom Windows command-line sanitizer (`quoteForCmd` plus the CR/LF refusal) or the intentional launcher-selection trust boundary, so the flows remain reported after `shell: false` is explicit. Every safety condition holds: argv array, no shell interpretation, shell use isolated to `.cmd`/`.bat`, CR/LF rejected, real Windows round-trip and side-effect tests green, launcher-path escaping reviewed and proven with metacharacters, and no test ever achieved execution outside the intended Harness launcher.

Per GitHub's documented workflow this is a false-positive classification — not "won't fix", which would concede an accepted vulnerability that the proof above shows is not there. No source-level suppression comment was added.

PR #134, opened 2026-09-02
feat(dshline): adopt Harness 0.1.2-alpha.5

riesbri:
## What this does

Moves dshline's one adopted Harness architecture forward and deletes what the old one needed.

| | from | to |
|---|---|---|
| version | `0.1.1-rc.2` | **`0.1.2-alpha.5`** |
| revision | `b150a551b8d465e31e418e1b2eaf5e79bbb7d28e` | **`db6bdc3576c2d4e7c965e8e3ed0c2a731eed87f5`** |

> This PR intentionally drops DeepSeek Harness `0.1.1-rc.2` compatibility. dshline tracks one adopted pre-1.0 Harness architecture at a time.

That is the policy working, not a regression. Nothing here is a shim, a version check, a feature probe, or an old/new union. Every `dsh-*` dependency, devDependency, and peerDependency is exactly `0.1.2-alpha.5` — no carets, no `||`.

Base is the released **0.15.0** tree plus the release-readiness guard (#135). Package versions are untouched at that baseline; Changesets owns the next one.

### The target moved while this was open

This PR was written against `0.1.2-alpha.4`. Upstream then cut `0.1.2-alpha.5` as a real release generation — GitHub Release `dsh-v0.1.2-alpha.5`, git tag, and a root manifest declaring it — so under *track aggressively* the PR advanced rather than merging a generation behind. alpha.4 support is not retained.

The adopted revision is the commit upstream tagged and released, `db6bdc35` (`release(dsh): 0.1.2-alpha.5`), not the later `49a606bc` sync-to-master: `HARNESS_TARGET` wants the immutable revision representing the generation, not whatever master happens to be. The release-PR merge `5a69ba1c` has a byte-identical tree, so the two would build the same thing; the tag is what upstream marks the release with, and it is what every previously adopted revision matched (`dsh-v0.1.1-rc.2` = `b150a551`, `dsh-v0.1.2-alpha.4` = `4e84901e`).

**alpha.4 → alpha.5 required no dshline change.** The only source changes upstream are in `session-projection-cache`, `storage`, `storage-domain`, `storage-json` and `tool-cordis` — none of which dshline depends on or governs. The projection *cache* is persistence-side derived data, distinct from the `session-projection` registry this PR moved `/plugins` onto, and that registry is untouched. The cache fix adds cross-version read compatibility whose failure modes are conservative in both directions: an old record missing lineage fields reads as unseeded and a seeded expectation then discards it to a cold rebuild, and an unparseable record is backed up and skipped. Neither can hand `stateOf()` a stale value.

## Session API migration

The `0.1.2` line removes the public `Session.events` array. Each read moved to the narrowest native API, never to a whole-log snapshot:

- **`context/model.ts`** — `eventAt()` point reads; the backward `callId` search keeps its early stop and is bounded by the newest wanted result. `ContextEntry.seq` is `SessionSeq`, so the meter's own branded node seq reaches `eventAt()` with no conversion.
- **`/work`** — the child's open-turn suffix is found by scanning back from `session.seq` and stopping at `inheritedEventCount`.
- **tests** — a whole-log assertion uses `snapshotEvents()`; a newest-event assertion uses `eventAt(seq - 1)`.

Arithmetic that leaves the branded domain re-enters through `SessionSeq(...)` under proven bounds. **Zero `as SessionSeq` / `as SessionLogOffset` casts**, and no `any`.

## `/work` lineage

The floor is `inheritedEventCount`, deliberately **not** `firstLiveSeq`. `firstLiveSeq` is the constructor-seed length, so on a cold-resumed child it covers that child's own work — using it would hide exactly the turn `/work` exists to describe. `inheritedEventCount` keeps the fork value across a resume, so resumed child-owned history stays visible while the parent's never leaks in. Harness refuses to fork inside an open turn, so the floor can never truncate a real one. Two regressions assert what is *not* read: no seq below the cut is fetched at all.

## `/plugins` ownership

Which preset a session runs is Harness's `agentPreset` Session projection; dshline's `resolveSessionPreset()` reverse scan is **deleted**, not translated. Switching is `agentPresets.select()` — Harness serializes per session, re-reads `turnBoundary` inside the switch, refuses a started session, recomposes, and records. dshline appends no `agent-preset/selected` and performs no write-path blank check.

`presetSwitchEligibility` survives as *presentation* eligibility only — deciding what to offer, reading the same projection Harness re-checks, so offer and refusal cannot drift. The one retained `recompose()` is the distinct case that is not a choice between presets: forcing a blank session to pick up a preset **file edit** live, where the id does not change. `window.ts`'s resume path reads the same projection, so there is one answer to "which preset does this session run".

## userQuestions

`registerQuestionAnswerer`, its runtime detection and its two casts are gone. One line onto the scoped `user-questions/request` waterfall, typed against `AskUserQuestionRequestEvent`. dshline's terminal answerer still never calls `next()`.

## Also deleted

The `subagent-model-selection-settings` row's `!!js` resolution probe. It existed only because rc.2 does not publish that subpath and the `0.1.2` line does — a capability probe whose answer is fixed by our own peer pin.

## Fixtures and `isSeeded`

Classified, not patched. Fresh root (skills probe: no seed, `isSeeded: false`); header-only presentation fixtures behind one shared factory defaulting to unseeded and explicitly independent of `parentSession`/`origin`, since Harness sets both on every subagent child regardless of a seed; and a new integration case that forks a **real** Session to cover inherited-child semantics with Harness's own constructor.

## Supply-chain policy

`minimumReleaseAge` is `180` (three hours), global and exception-free — no `minimumReleaseAgeExclude`, no scope carve-out, no CI override, no `trustPolicy` change. Verified against pnpm 11.22.0's own resolution (`Date.now() - minimumReleaseAge * 60 * 1e3`) that the unit is minutes. Three hours rather than a day because the `dsh-*` line is pinned exactly: a fresh upstream publish never enters the graph on its own, so the window brakes *what an adoption installs*, not *when* adoption may happen. alpha.5 was held by this window until it matured naturally; nothing was bypassed to adopt it.

## Changesets

`.changeset/thin-lions-repeat.md` stays **deleted** — 0.15.0 consumed it. The migration changeset is retained and renamed to name the generation actually adopted. No version chosen by hand.

## Release readiness is deliberately decoupled

> This PR may move development `main` ahead of the default published Harness channel. The release-readiness guard prevents a future dshline `latest` release until `@deepseek-ai/dsh@latest` exactly matches `HARNESS_TARGET`.

`@deepseek-ai/dsh@latest` is still `0.1.1-rc.2`, so after this merges the guard will correctly report **not ready**, and the generated Version Packages PR will stay red until DeepSeek promotes the adopted generation. That is the designed steady state, not a defect: development validates `HARNESS_TARGET`, publication waits. **`Release readiness · Harness latest` is skipped on this PR** — it runs only on the generated release branch, so no ordinary PR is ever gated on a pointer DeepSeek moves.

## Validation

Rebased onto the post-#135 `main` and revalidated end to end against the final alpha.5 target: build, typecheck, the full suite on Node 22.19/24/26, the Windows launcher hand-off, the exact adopted revision checked out from source and linked (`source typecheck` plus all nine capability probes), both published-consumer paths under a real pseudo-terminal, bilingual docs, advisories, CodeQL and workflow hardening.

## Deferred capabilities

`session.ownEvents()` / `isOwnSeq()`, ranged `snapshotEvents()`, and the `agent-preset/selected` Cordis event — none needed to replace old architecture, all natural follow-ups.


riesbri:
### The six red checks are one cause: the 24-hour release-age brake

Every failing job dies on the same step, before running anything:

```
[ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION] 51 lockfile entries failed verification:
  @deepseek-ai/dsh-brand@0.1.2-alpha.4 was published at 2026-09-01T15:46:36.001Z,
  within the minimumReleaseAge cutoff (2026-09-01T07:52:35.606Z)
```

`pnpm-workspace.yaml` sets `minimumReleaseAge: 1440`, and pnpm enforces it against **lockfile entries** — so `--frozen-lockfile` is checked too, and `Harness target` fails on its post-`link-harness` install for the same reason, before it can typecheck anything.

| job | failing step | cause |
|---|---|---|
| Core · Node 22.19 / 24 / 26 | `pnpm install --frozen-lockfile` | release age |
| Windows launcher | `pnpm install --frozen-lockfile` | release age |
| Harness published | `pnpm install --frozen-lockfile` | release age |
| Harness target | `link the built harness declarations` | release age |
| advisories (security.yml) | `pnpm install --frozen-lockfile` | release age |

Everything that needs no install is green: **Docs · EN/ZH pairing**, CodeQL, secrets, workflow hardening, new dependencies.

### Re-run after 2026-09-02T16:02Z

The newest package in the set is `@deepseek-ai/dsh@0.1.2-alpha.4` at `2026-09-01T16:01:23Z`, so the whole graph clears the 1440-minute window just after **16:02 UTC today**. A plain re-run of the workflow is all that is needed — no commit.

### Nothing was changed to get around it

No `minimumReleaseAge` bypass, exclusion, or CI flag was added; the lockfile is an ordinary resolution. This is the brake behaving exactly as [#133](https://github.com/riesbri/dshline/pull/133) described it — *"a legitimate security patch waits the same day, which is the trade"* — and adopting a generation on its publication day is the same trade. Weakening it to make a migration land faster would be the one thing it exists to prevent.

### What the failing lanes would have said

Each was run locally against this exact commit, with the brake overridden for the session only:

- Core: `build`, `typecheck`, **2477 tests** — green on the registry graph.
- **Harness target**: `4e84901e` cloned, installed, built, `link-harness --check` clean, `typecheck` exit 0, full suite green, and **all nine capability probes green** — `sessionQuery`, `jobs`, `subagents`, `sessionProjections`, `workflows`, `userQuestions`, `tokenMeter`, `compaction`, `skills`.
- **Harness published**: both consumer paths booted under a real pseudo-terminal — packed bundle beside `@deepseek-ai/dsh@0.1.2-alpha.4`, and first-run-from-nothing against the published pair.
- advisories: `pnpm audit --audit-level moderate` — no known vulnerabilities; `zizmor` — no findings.

riesbri:
### Superseded: the release-age window is now a policy decision, not a wait

My earlier comment said to re-run after 16:02 UTC. That is no longer the plan — `57c2de3` changes the repository policy instead:

```diff
-minimumReleaseAge: 1440
+minimumReleaseAge: 180
```

Three hours, not zero and not disabled. The control stays **global and exception-free**: no `minimumReleaseAgeExclude`, no `@deepseek-ai/*` carve-out, no CI override, no change to `trustPolicy` or frozen-lockfile verification. Confirmed against pnpm 11.22.0's own resolution (`Date.now() - minimumReleaseAge * 60 * 1e3`) that the unit is minutes.

The reasoning is specific to how this repo takes dependencies: every `dsh-*` package is pinned to the one exact generation in `HARNESS_TARGET`, so an upstream publish never enters the graph on its own — adoption is a deliberate commit. The window brakes *what an adoption installs*; it was never the thing deciding *when* adoption may happen, and a day-long one only delayed same-day Harness adoptions this project performs on purpose.

Stale references to the old 24-hour figure came along: `SECURITY.md` + its Chinese counterpart (pair re-recorded), the publish workflow's install comment, and one spec name. The spec's assertions were already duration-free — they check that no CI job reintroduces a bypass, which still holds and still passes.

### CI is now green on real, unmodified configuration

All thirteen checks pass; the two skips are the informational upstream lane and Scorecard.

The previously blocked jobs did not merely get past install — they executed their downstream work:

- **Core** on Node 22.19 / 24 / 26 — install, build, typecheck, **2477 passed / 22 skipped** on each.
- **Harness target** — checked out `4e84901e`, verified the generation, installed and built Harness, linked declarations, `source typecheck: PASS`, and **9/9 capability probes green**.
- **Harness published** — both real consumer paths ran: packed bundle beside `@deepseek-ai/dsh@0.1.2-alpha.4` (*profile loaded, banner, clean ctrl-d*), and first-run-from-nothing against the published pair (*profile created and installed, banner, clean ctrl-d*).
- **Windows launcher** — installed and ran the `.cmd` hand-off, 67 passed.
- **advisories** — installed and audited cleanly.

PR #152, opened 2026-09-04
feat(dshline): attach Harness-native raster images

riesbri:
## Problem

A terminal user could mention `@path`, but dshline had no way to send a real image to an image-capable model. Treating every path as an attachment would have changed existing source-file and directory semantics.

## Authority and design

The adopted Harness target (`0.1.2-rc.1`, `a66e4702`) supplies `ctx.fs` for bounded reads and `ctx.attachments` for deployment-resolved image limits, validation, normalization, durable batch admission, and opaque references. dshline owns only an explicit terminal gesture: `/image <path>` stages a PNG, JPEG, WebP, or GIF without I/O; sending an ordinary prompt reads through the current filesystem and submits one complete batch to Harness.

`@path` intentionally remains a textual file reference. Image drafts are session-local and process-local, while durable `ImageBlock` metadata is reconstructed from the session log. The renderer receives only presentation values; scrollback remains append-only. No provider names, storage paths, base64, or attachment IDs enter production presentation or persistence.

## Lifecycle and failure behavior

- Explicitly text-only model metadata rejects before I/O; unknown metadata remains Harness-owned.
- Registered commands receive images only after declaring `input.images`; the command registry performs its own durable admission.
- Failed, cancelled, or unavailable admission retains both the prompt and drafts. `ctrl-c` cancels one session-owned admission from read through dispatch. Late saves or command work cannot deliver into a disposed session.
- Filesystem errors use typed, path-free messages so an absolute input path is not echoed.
- Transcript history shows durable name/dimensions/size only, including image-only historical messages.

## Deliberate omissions

Harness currently has only durable image attachments. Arbitrary files remain unsupported until it publishes an equivalent generic file-attachment contract; source files and directories continue to be referenced textually and read by tools.

## Validation

- `pnpm build`
- `pnpm typecheck`
- `pnpm harness-target`
- `pnpm run verify-docs`
- `pnpm test` — 2,802 passing, 22 skipped
- `pnpm security` — audit and workflow lint clean
- `pnpm test:consumer` — packed plugin booted in a real pseudo-terminal and exited cleanly with `ctrl-d`
- `git diff --check`

The new interaction tests cover bounded reads, Unicode/CJK/spaces, limits, duplicates, service absence, command opt-in/error retention, text-only/unknown model metadata, cancellation during read and non-cancellable save, session transition, resume rendering, hostile filenames, and narrow hint shedding. I deliberately removed the post-save cancellation guard and confirmed the focused regression test failed before restoring it.

Manual note: the packaged pseudo-terminal smoke validates assembled startup, but intentionally has no model configured, so live model submission remains covered by the assembled Harness seam tests rather than claimed as a manual provider acceptance.


riesbri:
CI follow-up: the rerun passed Core on Node 22.19, 24, and 26, plus target, docs, Windows, security, CodeQL, and dependency checks.

The only blocking lane is `Harness published · consumer install path`. Its packed-current-commit boot passed; its separate `--bootstrap` phase installs the already-published `@dshline/dshline@0.15.0` from npm and fails before the PR artifact is loaded (`Cannot read properties of undefined (reading 'length')`). I reproduced that first-run failure locally as well (with a different local Harness-link resolution error), so this is a pre-existing published-release/bootstrap-path issue rather than an attachment implementation failure. The PR intentionally remains blocked pending that external release-path correction.

PR #156, opened 2026-09-04
chore(dshline): adopt Harness 0.1.3-alpha.1

riesbri:
Adopt DeepSeek Harness `0.1.3-alpha.1`.

| | version | revision |
|---|---|---|
| adopted | `0.1.2-rc.1` | `a66e4702047846cdaa10c66c9d3df3951f5ea70d` |
| candidate | `0.1.3-alpha.1` | `d347e703908d0406b7a7ef80e3a0e594d86b2215` |

Upstream release [`dsh-v0.1.3-alpha.1`](https://github.com/deepseek-ai/deepseek-harness/releases/tag/dsh-v0.1.3-alpha.1) · [compare](https://github.com/deepseek-ai/deepseek-harness/compare/a66e4702047846cdaa10c66c9d3df3951f5ea70d...d347e703908d0406b7a7ef80e3a0e594d86b2215) · newest official `dsh-v*` generation as of this push

> **This is no longer the mechanical pull request Harness-Sync opened.** That
> proposal was cut from `03d1633` — before 0.17.0, before `/image`, and before
> `dsh-attachment` entered the manifest — and CI answered *migration required*.
> The branch has been rebased onto post-0.17.0 `main` and carries the migration.

## Why a migration

Session format **v2**. Harness now stores ONE durable settlement per model
attempt — `assistant/message`, or `assistant/attempt` for a call that failed,
retried, or was cancelled without committing a message — each carrying its own
exact timed stream. The per-delta `assistant/chunk` log event is gone, and live
streaming is published as transient `agent/assistant-stream` frames instead.

dshline read four things from that deleted event: streamed text and reasoning,
time to first token, the `thinking`/`responding` status word, and the same word
for a Work child. All four move to the frame feed.

## What was deleted rather than adapted

- **`project()` loses its chunk branch.** It is the function the resume replay
  also runs, and streaming was the one thing in it the replay could never reach.
- **`isTranscriptEvent` loses the chunk filter.** Nothing is left in the log to
  skip; an `assistant/attempt` is not surface-eligible, replays like a tool call,
  and draws nothing — no special case.
- **`TurnTimer` keys spans by attempt, not by step.** A step that retried after a
  stream error no longer reports the gap between its two attempts as time the
  model spent producing text. The step-keyed version did.
- **`modelPhaseAfter` keeps only log boundaries**, plus `assistant/attempt`
  beside `assistant/message`. `modelPhaseAfterFrame` owns the transient half and
  clears on frame `end`, because an abandoned attempt publishes no durable event.
- **Work subscribes to its child's frames** with the identity filter it already
  used. Its open-turn seed loses streaming and cannot keep it: an in-flight
  attempt leaves no durable record until it settles.

Ordering is unchanged: `AssistantStreamAttempt.settle` appends the durable event
before publishing its `end` frame, so `assistant/message` still settles the
stream buffer where it always did.

## Command attachments

The registry renamed `input.images` → `input.attachments` and gave its
submission envelope a discriminant, its other arm being a staged file receipt.
`encodeCommandImages` stamps `type: 'image'`; dshline stages nothing and
produces no file arm. `/image` behaviour is unchanged.

## Compatibility deliberately NOT added

No version detection, no union of the two event shapes, no second reducer arm
for rc.1, no adapter preserving `assistant/chunk`. `@deepseek-ai/cordis` and
`@deepseek-ai/schemastery` floors are unchanged in this generation (4.0.2,
3.18.2), so no peer correction was needed.

## Validation

Re-run after rebasing onto `main` at `3466475`, which carries #158 (the
authorization seam this bundle now composes) and #159 (`/cache`). Numbers below
describe that rebased state, not the earlier pre-#158/#159 branch. Against the
exact candidate revision, source-linked with `tools/link-harness.mjs` the way
the `Harness target` lane links it:

- [x] `pnpm build`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 2926 passed / 22 skipped
- [x] `node tools/capability-report.mjs` — 12 / 12 seams, including `authorization`
      and `requestHeader`, the two the rebase brought in
- [x] `node tools/harness-target.mjs` and `--verify-source`
- [x] `node tools/lint-workflows.mjs`, `node tools/verify-translation-pairing.mjs`, `git diff --check`
- [x] Each new test broken on purpose and confirmed to fail by name

### Rebase conflicts and the stale prose they exposed

Two files conflicted, both resolved without touching production code.
`packages/dshline/package.json` took the union: `dsh-authorization` and
`dsh-credentials` arrived on `main` with #158 and both exist at
`0.1.3-alpha.1`, so both are kept at the adopted pin.
`docs/architecture.i18n.yaml` is the bilingual blob-hash record, re-recorded
through `pnpm run verify-docs --write`.

Adopting the generation then falsified prose written against `0.1.2-rc.1`, so
this branch corrects it and nothing more:

- **ROADMAP** claimed Harness exposes durable image attachments but no
  equivalent file contract. `0.1.3-alpha.1` publishes the generic half —
  `FileAttachmentRef`, `admitEncodedFile`, and a staged file-upload receipt arm
  on the command submission envelope — so what is missing is a dshline staging
  gesture, not a contract. **No file-attachment implementation is in this pull
  request**; `@path` stays textual.
- **`docs/usage`** and **`docs/design`** named the removed `input.images` flag;
  both now name `input.attachments`.
- **`docs/usage`**'s illustrative `/setup` output presented `0.1.2-rc.1` as the
  adopted generation. The healthy row now reads `0.1.3-alpha.1`, and the
  mismatch example is inverted rather than rewritten — it now shows the
  published `0.1.2-rc.1` installed against an adopted `0.1.3-alpha.1`, which is
  the state a person actually hits today.

`0.1.2-rc.1` survives on purpose in `tests/setup-model.spec.ts`: those literals
are `compareGenerations` fixtures exercising match/mismatch/unknown, not a
record of what this build adopts.

## ⚠️ Do not merge yet — the lockfile is missing and cannot be written

`0.1.3-alpha.1` **is not published to npm at all** (not a dist-tag lag: no
`dsh-*` package carries the version). `pnpm install` cannot resolve a single
spec, so no lockfile can be generated, and every registry lane — **Core**,
**Windows launcher**, **advisories** — stays red on
`ERR_PNPM_OUTDATED_LOCKFILE`. **Harness target**, the lane that answers whether
dshline works against this architecture, is what this push turns green.

When upstream publishes, one `pnpm install` commit finishes the adoption.

### Why the original proposal had no lockfile either

`harness-sync.yml` runs `pnpm install … 2>&1 | tee install.log` under
`bash -e` with no `pipefail`, so the step took `tee`'s exit status. The install
had failed with `ERR_PNPM_NO_MATCHING_VERSION` (the release was nine minutes
old and not yet on npm), the quarantine classifier never ran because the step
was reported as a success, and the pull request was opened with manifests and
no lockfile. That is a separate one-line workflow fix, not part of this
adoption.


riesbri:
## CI on `889fba2`

| lane | result | why |
|---|---|---|
| **Harness target · adopted source revision** | ✅ | `source typecheck: PASS`, all ten capability seams `✓`. This is the lane that answers *does dshline work against this architecture*. It was **red** on the generated commit (`assistant/chunk` removed from `SessionEventMap`). |
| Harness published · consumer install path | ✅ | Reports publication lag and validates nothing — `0.1.3-alpha.1` is not on npm. |
| Docs · EN/ZH pairing | ✅ | |
| Core · Node 22.19 / 24 / 26 | ❌ | `ERR_PNPM_OUTDATED_LOCKFILE` |
| Windows launcher · npm .cmd shim | ❌ | `ERR_PNPM_OUTDATED_LOCKFILE` |
| Security → advisories | ❌ | `ERR_PNPM_OUTDATED_LOCKFILE` |
| Security → codeql / secrets / new dependencies / workflow hardening | ✅ | |

Every red lane is the same single cause, and it is not a compatibility question:
`pnpm install` cannot resolve `@deepseek-ai/dsh-*@0.1.3-alpha.1` because no such
version exists on the registry, so no lockfile can be written for this
generation. Verified directly:

```
$ npm view @deepseek-ai/dsh-agent@0.1.3-alpha.1 version
npm error 404 No match found for version 0.1.3-alpha.1
$ npm view @deepseek-ai/dsh dist-tags
{ latest: '0.1.2-rc.1', alpha: '0.1.2-alpha.5', next: '0.1.2-rc.1' }
```

**Hold.** When upstream publishes the generation, `pnpm install` on this branch
writes the lockfile and every remaining lane goes green with no source change.
Release stays gated separately on `@deepseek-ai/dsh@latest` reaching
`0.1.3-alpha.1`, per `HARNESS_TARGET`'s release-channel rule.

PR #158, opened 2026-09-05
feat(dshline): guide a fresh install to a working model, and compose the authorization seam

riesbri:
A new user can reach a fully installed, correctly booted dshline and still have nothing to talk to. `/connect` and `/model` both exist and both work; nothing points at them, and the composer they land in instead cannot send a turn without a route.

Worse, the path a beginner is most likely to take — sign in with an account — was not merely undiscoverable. **It was unavailable.**

## 1. Account sign-in was dead in every stock profile

At the adopted generation **no shipped Harness bundle mounts `@deepseek-ai/dsh-authorization`** — not `dsh-base`, not `web-app`, not `headless`, not `acp-app`, not the `dsh` app. Checked with `git grep dsh-authorization a66e470` across every `*.yml`/`*.yaml`/`*.ts`/`*.json`; it hits only `llm-pi-ai`'s source, its tests, and lockfiles.

`registerPiAiFlows` is scoped with `ctx.inject(['authorization'], …)`, so a composition without that row registers **no flows at all**: `/connect`'s Sign-ins section was permanently empty and an account-authenticated provider was unreachable from a terminal.

This bundle's `cordis.patch.yml` now inserts the row, exactly as it already inserts `session-stats` and for the reason that comment gives — a host-plane seam this surface reads that the base does not carry. That is composition, which is what a bundle is for.

**The obvious alternative is wrong.** `dsh-authorization` declares no `dsh.bundle`, so `dsh plugin add` would install it as a plain dependency that composes nothing — the "installed but inert" state `/profiles` already reports — and finishing the job would mean dshline writing a composition row into the profile's own `cordis.patch.yml`, a patch layer no Harness mutation API owns. A first-run installer for a capability a bundle can simply compose is a second lifecycle where none is needed.

Because the row is now ours to mount, its shape is our compatibility problem too: a new capability probe mounts the real service over a real abstract `CredentialProvider` subclass, registered in `tools/capability-probes.mjs`.

## 2. Signing in is not activating a route

A credential record and a settings profile are separate writes to separate stores, and both are required:

```
ctx.credentials   the RECORD an authorization flow commits   llm-pi-ai/openai
ctx.settings      the PROFILE that registers a route         llm-pi-ai.providers.openai
```

`registerPiAiFlows` offers a login for every installed catalog provider "independent of the route set"; the adapter registers routes only for the profiles settings supply. Both halves are right, and together they produce the failure this PR exists for: you sign in, the flow reports success, and `/model` still says nothing is configured.

`connect/activation.ts` is the missing sentence and nothing more:

```
Signed in · ChatGPT (Codex)

  This account is authorized, but the openai model route is not active,
  so /model still offers nothing from it.

❯ Activate the openai route   Writes a profile in llm-pi-ai so the adapter registers it
  Not now                     Writes nothing
```

**A successful authentication is not consent to change provider configuration.** `Not now`, `esc`, and a closed browser all leave settings untouched. The write is the *existing* `activateRoute` through `ctx.settings.mutate` — no new mutation surface, no YAML editing.

Two details that are not details:

- The offer is judged from a **fresh** `ConnectCatalog.reread()`, never the snapshot the sign-in began under. `activateRoute` writes an empty profile at the route's path, so a stale "dormant" would replace configuration something else added in the meantime.
- The sign-in ↔ route link lives in `connect/pi-ai.ts`, **not** `model.ts`. `model.ts` still refuses the join in general and is right to. What changed is the claim, not the refusal: `dsh-llm-pi-ai` documents the correspondence for *itself* on both sides — `recordKeyFor`'s parameter doc calls its id *"pi-ai's own provider id, which is also the harness route key"*, and `directoryEntries` publishes that id at `providers.<id>`. It is verified, not assumed: a key resolves only against routes the directory actually published, at the address it publishes them at.

## 3. Setup is a conductor, not a wizard

A stock fresh install, captured from a packed build of this branch booted beside the real published launcher:

```
$ dshline

Setup

· Node       24.18.0
· dshline    0.17.0
✓ Harness    0.1.2-rc.1
✓ Profile    dshline
✓ Connecting API key · account sign-in
✓ Models     deepseek-official/deepseek-v4-flash · 1 route active · deepseek-official
⚠ Provider   deepseek-official needs a credential · DEEPSEEK_API_KEY is not set
  Connect a provider below to sign in or store a key, or choose a model on another route.

❯ Connect a provider
  Choose a model on another route
  Not now
```

Harness ships a default model *and* registers its route before any key exists, so `Models` is genuinely healthy here — the model is selected and the route is live. The credential is what is missing, which is why it gets its own row and why connecting leads the steps rather than choosing a model. With nothing configured at all the report instead warns on `Models` and leads with `Connect a provider` → `/connect` → sign in → activate → `/model` → `✓ Ready.`

`src/setup/` runs inside the composed Host, because every fact it needs (`ctx.llm`, `ctx.settings`, `ctx.credentials`, `ctx.authorization`) exists only there. Reaching for them from `bin/dshline.mjs` would make the wrapper a second reader of Harness state outside Harness — and `dshline --setup` already means "install this package into the profile", so it keeps that meaning.

It opens by itself when this launch would otherwise reach a composer that cannot send a turn. Four states, checked in this order:

| | |
|---|---|
| `no-route` | no adapter registered any provider route |
| `no-selection` | routes exist, but nothing resolved a model selection |
| `unregistered-selection` | a selection exists naming a route no adapter registered — a remembered default whose provider left the profile |
| `credential-missing` | the topology is complete, and the selected route's credential is one Harness **positively** reports absent |

The first three are synchronous registry and selection reads and settle most launches on their own. Only a launch that passes all three pays the fourth, and that one is `/connect`'s own `readRouteReadiness` over the **selected route only** — no `listModels`, no discovery, no network. Measured: `settings.describe()` is in-memory at ~1.2 ms median for 35 namespaces, and `credentials-local.describe()` is a map lookup over a snapshot loaded at mount.

**`unknown` is never a fault.** A route naming no credential reference is not misconfigured — it is authenticating through an account sign-in or provider-native discovery, and a signed-in `llm-pi-ai` route stores no `apiKeyEnv` (that field has no schema default, unlike `llm-deepseek`'s). A store that cannot answer is unread, not unset. Neither opens setup, and nor does an absent seam.

The fourth state is why a stock first install no longer looks healthy: `dsh-base` composes a default selection (`deepseek-official/deepseek-v4-flash`) and `llm-deepseek` calls `registerAdapter` unconditionally, so all three topology checks pass while the first prompt fails `MISSING_CREDENTIAL`.

There is deliberately no first-run marker anywhere — a stored "already set up" flag is duplicated state that can disagree with the configuration it describes. `/setup` reopens the flow on demand.

Each step hands off to the browser that already owns it, so there is **no second route editor, no second model catalogue, and no state machine**. The report is committed to scrollback rather than drawn in an overlay: version numbers and the sentence naming why there is no model are the output most worth keeping, and a bounded live region would scroll them away. The only live surface it uses is `promptSelect`, already bounded and resize-safe.

### Compatibility is reported, never gated

Harness publishes no runtime version service at this generation (checked `dsh-brand`, `dsh-home-paths`, the loader; `dsh-plugin-package-inventory-deepseek` builds a package list only as request metadata). So the evidence is manifests, read through the machinery `/profiles` already owns: the adopted generation is this package's own `dsh-*` peer pin — which `tools/harness-target.mjs` proves is `HARNESS_TARGET.version`, so **no example version is written down anywhere** — and the installed one is the `dsh-base` the running profile composes.

Three rules, each a refusal:

- **A mark is a claim, so an unknown gets no mark.** Either side unreadable is `·` — never "incompatible", never "fine".
- **A mismatch names both versions and one deterministic repair.** It tells the reader how to install the exact Harness generation this dshline targets. Moving dshline instead is described only conditionally, because this report cannot know which released dshline targets an arbitrary installed generation — establishing that would mean resolving releases against their peer pins.
- **It never offers "continue anyway".** By the time the comparison can be drawn, both halves have already booted together far enough to draw it; a genuinely incompatible pair fails earlier and louder in the Loader, which is Harness's diagnosis to give.

Node is reported without a verdict for the same reason in miniature: this process is already running on it, and turning `engines` into a pass/fail means evaluating a semver range.

## Harness owns / dshline owns

Harness owns the flows, the seam, the credential store, the settings document and its revisions, route registration, the model catalogue, package lifecycle, and the composed tree. dshline owns terminal presentation, an *ordering* of existing surfaces, one domain-scoped reading of `llm-pi-ai`'s own documented identity, and one bundle row.

## Verified

Full suite **2902 passed / 22 skipped** across all three projects; `pnpm build`, `pnpm typecheck`, `node tools/harness-target.mjs`, `pnpm run verify-docs`, and `pnpm security` all clean.

Deliberate mistakes, each confirmed to fail the test named for it:

| Broken on purpose | Fails |
|---|---|
| removed the consent gate in `offerRouteActivation` | "writes nothing when the reader chooses Not now", "…when the reader dismisses the question" |
| judged from `row.route` instead of the re-read | "asks nothing and writes nothing when the route is already active", "judges the offer from a fresh reading…" |
| treated a dismissed step picker as "not skip" | the three flow tests that leave setup (time out) |

The composition was verified end to end by running the real `applyEntryPatches` over the real `dsh-base` patch and this one: 85 rows in, 91 out, `authorization -> @deepseek-ai/dsh-authorization` present, base carried no such id, all 25 disables intact. The package resolves at `0.1.2-rc.1` from both anchors `resolveBundleDir` uses.

A packed build of this branch was also booted in a fresh profile beside the real published launcher, under a pseudo-terminal, and rendered:

```
✓ Connecting API key · account sign-in
```

That half of the row appears only when `ctx.authorization` is mounted **and** a plugin has registered flows into it, so the composed authorization seam is live and `llm-pi-ai` registered its account sign-ins in a real profile.

## Deliberately not implemented

- No install-the-plugin flow (see §1), no plugin marketplace, no provider registry.
- No "Update Harness" / "Continue anyway" / "Exit" on a version mismatch.
- No Node verdict.
- No profile changes, no `/doctor`, no background polling, no duplicated credential/model/profile state.

## Follow-ups (not in this PR)

- **`/doctor`: recommend not adding it.** The report `/setup` now commits *is* the read-only version, minus the sections that would duplicate `/profiles` and `/connect`. If wanted anyway, make it an alias that runs the report and skips the step picker — `reportLines(await gatherSetupFacts(...))` is already the whole implementation.
- **Profiles:** `dshline --setup` is the documented retry for a broken profile but is undiscoverable from inside a session. Worth one line in `/profiles` naming the command for the current profile.

## Docs & changeset

`docs/usage`, `docs/architecture`, and `docs/install` updated in **both languages** with `.i18n.yaml` re-recorded (`9 bilingual pairs consistent`). One `minor` changeset added.


riesbri:
-

riesbri:
-

PR #177, opened 2026-09-07
feat(dshline): render the structured metadata tool views already publish

riesbri:
## What

Render the structured metadata the pinned Harness presentation contract already publishes, which the cards currently omit — without special-casing any tool name and without new domain state.

- **Declared locations (`FileLocation[]`)** — a `generic` or `diff` call view's `locations` now render as one `path` (or `path:line`) row under the call title, in the view's declared order. The list shares the call card's row budget, hides entirely at `hidden`, and arms ctrl+o when elided (`InspectableToolCall` carries the locations; the inspector re-renders them at its own budget). Locations are **never** inferred from arguments or diffs when the field is absent.
- **Web search** — the provider's `answer` renders as wrapped card content above the source list; each source now shows its snippet and publishedAt beneath its title/url; publishedAt is displayed safely without date parsing or reformatting. Source order, the count, and the `+` capped indicator are unchanged. One row budget is spent across the answer and all sources, the same rule a bulk diff already follows.
- **Web fetch** — the card opens with a retrieval summary from the structured view — url, HTTP status, and the view's own `truncated` flag (`⎿ https://… · 200 · truncated`, marker in the `warning` role) — with the fetched body below at the existing budgets. Truncation is never recomputed from content length, and the model-facing `Fetched …` envelope is rendered as content, never parsed for metadata.

## Why

The metadata rode on every `tool/call` and `tool/result` event and never reached the screen: a reader watching a multi-file call saw only its title, a search showed titles beside urls and nothing the provider said about them, and a fetch was indistinguishable from any raw text dump.

The obvious shortcut — routing every non-`search` web arm to the new fetch renderer — would have broken forward compatibility: an unknown future arm must degrade to raw content rather than read fields it does not carry. The dispatch checks for the arms it knows explicitly.

## Forward compatibility

At compile time the pinned union types are used naturally; at runtime every unknown `ToolCallView`/`ToolResultView` arm keeps the sanctioned raw-content fallback. A test pins this for the web union.

## Tests

25 new tests in `cards.spec.ts` (23 across three new describes, plus a web-union fallback pin). Each new behavior was broken on purpose and confirmed to fail tests by name: dropping the `:line` suffix (3 failures), ignoring `answer` (1), hard-disabling the fetch `truncated` marker (4).

## Changeset

`patch` for `@dshline/dshline` (presentation-only, per the precedent of prior card-presentation changes).

### DIFF
--- a/.changeset/fluffy-moons-render.md
+++ b/.changeset/fluffy-moons-render.md
@@ -1,0 +1,5 @@
+---
+'@dshline/dshline': patch
+---
+
+Render more of the structured metadata tools already publish through their presentation views. A call card now lists the files its view declared via `locations` (with `:line` when the view focuses one), bounded by the same row budget as the body and reachable through the inspector when elided. A web search card shows the provider's `answer` above its sources, and each source now carries its `snippet` and `publishedAt` alongside the title and url. A web fetch card opens with a retrieval summary — url, HTTP status, and the view's own `truncated` flag — instead of showing only the fetched text.
--- a/packages/dshline/src/cards.ts
+++ b/packages/dshline/src/cards.ts
@@ -21,6 +21,7 @@ import type {
   DiffCallView,
   DiffResultView,
   FileDiff,
+  FileLocation,
   ReadResultView,
   SearchResultView,
   TerminalCallView,
@@ -28,12 +29,16 @@ import type {
   ToolCallView,
   ToolDefinition,
   ToolResultView,
+  WebFetchResultView,
   WebResultView,
+  WebSearchResultView,
+  WebSource,
 } from '@deepseek-ai/dsh-tools'
 import { diffRows } from './diff.ts'
 import {
   box,
   BOX_CHROME_COLUMNS,
+  displayWidth,
   escapeControls,
   hangingIndent,
   paint,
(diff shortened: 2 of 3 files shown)


riesbri:
Addressed the row-budget finding in 6f98dd7.

**The fix.** The generic call's locations and content now spend one shared allowance through a single `callSections` helper: locations draw first, the content receives `budget - drawn`, and each section reports what the shared budget hid from it with its own elision marker. Both the scrollback card and the inspector render through that helper, so the inspector reconstructs both semantic sections at its own budget.

**Invariants held:** locations remain before content; `hidden` draws neither; compact/full/inspect keep their existing numeric budgets; either section being cut still arms ctrl+o. Structurally, `limit` now takes a raw budget and `body` accepts the leftover allowance as an override — one budget mechanic, no second system. A diff call's locations remain its whole body with the full allowance; web rendering, result cards, and tool-name routing are untouched.

**Regression.** "spends ONE row budget across a call's locations and content together" — five locations plus ten content lines at compact: exactly six secondary rows (five locations, one content line), `… 9 more lines · ctrl+o view`, and the inspector exposes both `file4` and `echoed 9`. A companion test checks full detail through the same helper (fifteen rows, no elision, no arm). Verified the regression fails against the old two-budget code (eleven secondary rows) and passes with the fix.

`pnpm typecheck` and the full suite (3,010 passed) green.

PR #186, opened 2026-09-09
fix(dshline): hold a foreign raw write off the live region's terminal

riesbri:
Spawning a subagent printed dshline's root chrome — `╭─ dshline ─… ─╮` — into
native scrollback a second time, and once more per commit after that. Not the
work/subagent projection, not `activeWork`, not `/work`.

> **This is a temporary compatibility shim for an upstream defect, not a dshline
> abstraction.** It is registered in `HARNESS_COMPAT`, so the next Harness
> generation cannot be adopted without someone deciding whether it is still
> needed. The upstream issue draft is
> [in a comment below](#issuecomment-5600886712).

## The invariant

`packages/renderer/src/screen.ts` is correct only because it is the sole writer:
it remembers the live region's height and where it left the cursor, and every
redraw climbs that remembered geometry to erase the frame before drawing the
next one. Its own module doc states the rule — *"the live region must be the
last thing on screen, so every write goes through this class."*

A write it did not issue scrolls the screen out from under that geometry. With
`liveRows.length = 5` and `cursor = { row: 2 }` (the composer's input row),
`eraseLive()` descends `bottom - cursor.row` = 2, `\r`, then climbs `bottom` =
4 — so after a foreign two-line write the origin lands **two rows below** the
real region top, and `CSI 0J` clears from there. The separator and the top
border sit above the erase origin, are never erased again, and become durable
output.

## The foreign writer

A subagent backend in the generation named by `HARNESS_TARGET` forwards a
delegated child process's stderr with `writeFileSync(process.stderr.fd, bytes)`,
unconditionally, for the whole life of the delegation. On an interactive launch
descriptor 2 is the terminal dshline is drawing on, so the child's startup
diagnostics land inside the composer frame — which is why this appears the
moment a subagent starts, and why `/work` is irrelevant. It is the only
direct-descriptor terminal writer in the whole installed dependency tree
(`grep -rn "writeFileSync(process" node_modules/@deepseek-ai/*/lib/*.js` → one
hit). No dshline runtime file names that backend or branches on a provider: what
is contained is a **write shape**.

## The containment

Patching `process.stderr.write` would not catch it: the forward never touches
the stream. What it does read, on every write, is `process.stderr.fd` — a
writable, configurable own property. So that is what moves. While the window
owns the terminal, the descriptor that property reports is an `os.devNull`
handle:

```ts
ctx.effect(() => holdStderrOffTerminal(), 'dshline: foreign stderr writes')
```

Taken with terminal ownership in `createWindow` and released with it, exactly
once.

### What it knows before it engages

`isTTY` on both streams does **not** establish that stdout and stderr are the
same terminal, and the first version of this check claimed it did. It answers a
weaker question — each descriptor is *a* terminal — and two terminals are
exactly the case where moving stderr would be a pointless mutation of a process
global.

`rawStderrReach` now asks the answerable question. A terminal is a character
device, and `fstat`'s `rdev` is the device a handle is on, so two descriptors on
one terminal report one non-zero `rdev` and descriptors on two terminals report
two. Three answers, and the third is the honest absence of one:

| | meaning | behaviour |
|---|---|---|
| `reaches` | proven same terminal device | hold |
| `cannot-reach` | not a terminal, or a proven *different* terminal | leave completely alone |
| `unknown` | platform cannot say (a Windows console reports `rdev` 0; a stream may expose no descriptor) | hold |

Holding on `unknown` is the conservative asymmetry made explicit: being wrong
that way drops a raw diagnostic on a terminal nobody was drawing on, being wrong
the other way corrupts a transcript the reader cannot repair. A `2>log` run, a
piped stderr and a proven second terminal all keep their descriptor as found.

### The Node behaviour it depends on, pinned by real processes

Whether substituting the descriptor is safe is a question about Node, and a
stubbed `WriteStream` only proves what the stub does. `stderr-descriptor.spec.ts`
drives real `node` children with real descriptors and reads the files afterwards:

- a raw `writeFileSync(process.stderr.fd, …)` **follows** the substitution, and
  the substitution reverses exactly;
- a **socket-backed** `process.stderr` does **not** re-resolve the property, so
  the ordinary stream path keeps its own destination. The case driven is a
  **pipe**; production is a terminal, the same stream family
  (`tty.WriteStream extends net.Socket`), but this suite cannot allocate a pty
  — so that step is named as an **inference** in `src/stderr.ts` rather than
  presented as a measurement;
- a **file-backed** `process.stderr` is `SyncWriteStream` and *does* re-resolve
  `this.fd` on every write, so holding one would capture ordinary stream writes
  too. This was found by writing the test and is now the **stronger** reason the
  shim refuses a non-terminal stderr — the first reason was only that no frame
  could be corrupted. The refusal is load-bearing, so the reason is recorded as
  a test rather than as a comment.

### What it deliberately is not

No capture, tee, or logging sink. Harness owns Host diagnostics and subagent
failure reporting, and a second authority over the same bytes in the frontend
would be worse than dropping them. A raw-descriptor writer's output is dropped
while the window draws: that costs no failure reporting — a run's own failure
reaches the transcript through the subagent lifecycle, never through descriptor
2 — and the bytes were unreadable anyway, because they were landing on top of
the frame they corrupted.

## Making sure it does not become permanent

Prose does not fail a build. A workaround nobody is made to revisit is a
workaround that becomes permanent and is eventually read as a design decision.

**`HARNESS_COMPAT`** (new, at the repository root beside `HARNESS_TARGET`) lists
each temporary workaround with the generation it was last confirmed to still be
needed against:

```
shim packages/dshline/src/stderr.ts 0.1.5-alpha.1
```

`node tools/harness-target.mjs` — the coherence check the Harness-Sync adoption
proposal and every release **already** run — now fails while a record names any
generation other than the adopted one, and the blocking `Harness target` lane
runs it too, so a hand-edited `HARNESS_TARGET` is caught there rather than at
release. A record whose module no longer exists fails as well, so the register
cannot outlive what it describes.

The resulting migration flow is exactly the one asked for:

```
HARNESS_TARGET bumped
→ node tools/harness-target.mjs fails: "stderr.ts last confirmed against 0.1.5-alpha.1"
→ adopter reads the module header and checks the upstream checkout the migration already has
→ still broken  → bump the record, deliberately, in the migration commit
→ fixed         → delete stderr.ts, its ctx.effect, its specs, and the record
```

Deliberate non-goals, per review:

- **No upstream parsing.** The check is a string comparison between two files in
  this repository, so nothing couples a build to a backend's internal layout and
  a refactor that fixed nothing cannot silently satisfy it. Confirming the
  behaviour is the adopter's job, once per generation.
- **No parallel mechanism.** It is consumed by the existing adoption tooling and
  the existing checks; `HARNESS_TARGET`'s own header points at it.
- **Kept out of `HARNESS_TARGET`.** That file's format is two `key value` lines
  precisely so a migration can edit it without understanding a parser, and
  records carrying a path would erode that.
- **No runtime dependency.** Nothing in dshline's runtime imports
  `dsh-subagent-codex` or branches on a provider; `work.spec.ts`'s
  provider-neutrality gate still holds.

## What was ruled out, by probe rather than by reading

- **Live-region overflow.** Property-probed the real `TuiSlots.compose` with the
  real views across a wide width × height matrix and every composer state:
  physical rows never exceed the terminal at ordinary widths. (The narrow-width
  overflow that probe *did* find is #187, and is not this bug.)
- **Every overlay** already carries a `physicalRows(...) <= terminalRows`
  backstop, work's included, and its fallback is bounded.
- **`Screen.commit` arithmetic**, at exactly-full-height and over-height
  regions: the commit/redraw ordering is correct.
- **The subagent projection path** only calls `invalidate()`, and its
  `session/event`, `agent/assistant-stream` and `agent/status` listeners are all
  filtered by exact session or agent identity.

## Tests

`packages/dshline/tests/foreign-writes.spec.ts` (13) — three separate claims:

- a **hazard guard** on a real terminal emulator asserting that a raw write into
  the region *does* make root chrome durable. If it ever reports one row, a
  foreign write stopped being able to displace the region and this containment
  is no longer load-bearing — the suite says so rather than passing quietly.
- forty turns of the two writes the window actually makes (a synchronous commit
  per session event, a coalesced repaint after it) leave the terminal holding
  root chrome **exactly once**.
- `rawStderrReach` against **real character devices**: two handles on one device
  prove `reaches`; two different devices prove `cannot-reach` and are left
  untouched; a non-terminal at either end settles it without a stat; a regular
  file claiming to be a terminal and a stream with no descriptor both answer
  `unknown`. Plus policy: release is idempotent, an accessor-backed `fd` is
  refused rather than redefined, and `write` is never touched.

`packages/dshline/tests/stderr-descriptor.spec.ts` (2) — the Node behaviours
above, in real child processes.

`tools/harness-target.spec.mjs` (+8) — the register: record parsing and its
refusals, an absent register treated as empty (so a released tag still checks),
a record naming another generation fails, one naming the adopted generation
passes, a record whose module is gone fails, the report names the *decision*
rather than only the failure, the committed register agrees with the committed
target, and every recorded shim says `TEMPORARY` in its own first lines and
points back at `HARNESS_COMPAT`.

```
Test Files  168 passed | 1 skipped (169)
Tests       3243 passed | 22 skipped (3265)
pnpm typecheck / pnpm build   clean
git diff --check              clean
node tools/lint-workflows.mjs no findings
```

Independent of #187 — the two touch no common file.

### DIFF
--- a/.changeset/hold-foreign-stderr-off-the-live-region.md
+++ b/.changeset/hold-foreign-stderr-off-the-live-region.md
@@ -1,0 +1,34 @@
+---
+'@dshline/dshline': patch
+---
+
+Keep a foreign raw write off the terminal the live region owns, so spawning a
+subagent stops printing the root chrome into scrollback a second time.
+
+`Screen` is correct only because it is the sole writer: it remembers the live
+region's height and where it left the cursor, and every redraw climbs that
+remembered geometry to erase the frame before drawing the next one. A write it
+did not issue scrolls the screen out from under that geometry, and from then on
+the erase starts below the frame's first rows instead of above them — so the
+blank separator and `╭─ dshline ─… ─╮` are never erased again, scroll up as
+ordinary output, and stay in native scrollback for good. One more copy lands
+with every commit that follows.
+
+A subagent backend in the adopted Harness generation forwards a delegated child
+process's stderr with `writeFileSync(process.stderr.fd, bytes)` for the whole
+life of the delegation. On an interactive launch descriptor 2 is the terminal
+dshline draws on, which is why the duplicate appears the moment a subagent
+starts and has nothing to do with whether `/work` is open. Patching
+`process.stderr.write` would not catch it — the forward never touches the
+stream — but it reads `process.stderr.fd` on every write, so that is what moves:
+while the window owns the terminal, the descriptor that property reports is a
+writable hole rather than the terminal. `process.stderr.write` is left alone, so
+dshline's own boot-failure reporting and every ordinary stream writer are
+unchanged, and a stderr that is not a terminal (`2>log`) is not touched at all.
+
+The cost is stated rather than hidden: a raw-descriptor writer's bytes are now
+dropped instead of overwriting the input frame. That loses no failure reporting
+— a backend's run failures reach the transcript through the subagent lifecycle,
+not through descriptor 2 — and the bytes being dropped were unreadable anyway.
+The real fix belongs upstream, in a backend that should not write to a
+frontend's terminal; this keeps the frame correct until it does.
--- a/.changeset/hold-foreign-stderr-off-the-live-region.md
+++ b/.changeset/hold-foreign-stderr-off-the-live-region.md
@@ -14,21 +14,47 @@ blank separator and `╭─ dshline ─… ─╮` are never erased again, scrol
 ordinary output, and stay in native scrollback for good. One more copy lands
 with every commit that follows.
 
-A subagent backend in the adopted Harness generation forwards a delegated child
-process's stderr with `writeFileSync(process.stderr.fd, bytes)` for the whole
-life of the delegation. On an interactive launch descriptor 2 is the terminal
-dshline draws on, which is why the duplicate appears the moment a subagent
-starts and has nothing to do with whether `/work` is open. Patching
-`process.stderr.write` would not catch it — the forward never touches the
-stream — but it reads `process.stderr.fd` on every write, so that is what moves:
-while the window owns the terminal, the descriptor that property reports is a
-writable hole rather than the terminal. `process.stderr.write` is left alone, so
-dshline's own boot-failure reporting and every ordinary stream writer are
-unchanged, and a stderr that is not a terminal (`2>log`) is not touched at all.
+A subagent backend in the generation named by `HARNESS_TARGET`
+(`@deepseek-ai/dsh-subagent-codex@0.1.5-alpha.1`, `startCodexRun`'s stderr
+forward) writes a delegated child process's stderr straight to descriptor 2 —
+`writeFileSync(process.stderr.fd, bytes)` — unconditionally, for the whole life
+of the run. On an interactive launch descriptor 2 is the terminal dshline draws
+on, which is why the duplicate appears the moment a subagent starts and has
+nothing to do with whether `/work` is open. Nothing in dshline's runtime names
+that backend or branches on a provider: what is contained is a write shape.
 
-The cost is stated rather than hidden: a raw-descriptor writer's bytes are now
-dropped instead of overwriting the input frame. That loses no failure reporting
-— a backend's run failures reach the transcript through the subagent lifecycle,
-not through descriptor 2 — and the bytes being dropped were unreadable anyway.
-The real fix belongs upstream, in a backend that should not write to a
-frontend's terminal; this keeps the frame correct until it does.
+**This is a temporary compatibility shim, not a dshline abstraction.** The real
+fix is upstream — a delegated child's diagnostics belong on a Host-owned
+diagnostic seam, not on a frontend's terminal. `src/stderr.ts` carries the
+removal condition in its own header: it is deleted, with its wiring and its two
+specs, when `HARNESS_TARGET` advances to a generation whose subagent backends no
+longer write raw child diagnostics to the frontend's terminal.
+
+Patching `process.stderr.write` would not catch the forward — it never touches
+the stream — but it reads `process.stderr.fd` on every write, so that is what
+moves: while the window owns the terminal, the descriptor that property reports
+is a writable hole. Whether that is safe depends on two Node behaviours, and
+both are now pinned by real child processes rather than by stubs: a raw
+`writeFileSync` follows the substituted descriptor, and a socket-backed
+`process.stderr` — a pipe in the test, a terminal in production, both writing
+through a libuv handle opened once — does not, so the ordinary stream path keeps
+reaching the terminal. A FILE-backed `process.stderr` is `SyncWriteStream` and
+*does* re-resolve the property on every write, which is recorded as the stronger
+reason the shim refuses to hold a stderr that is not a terminal.
+
+That refusal is now a real device-identity test rather than an inference.
+`isTTY` on both streams does not establish that stdout and stderr are the same
+terminal — the previous version of this check claimed it did — so
+`rawStderrReach` stats both descriptors: two handles on one terminal report one
+non-zero `rdev`, and two terminals report two. It answers `reaches`,
+`cannot-reach`, or `unknown`, and holds on the first and the last. A Windows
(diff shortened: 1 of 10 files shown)


riesbri:
## Upstream issue draft — DeepSeek Harness

Filed here for review before it goes upstream. dshline's side of this is a
temporary shim with a written removal condition (`src/stderr.ts`), so this is
the change that actually retires it.

---

**Title:** `subagent-codex` writes a delegated child's stderr to the host's
terminal, corrupting any frontend that owns the screen

**Affects:** `@deepseek-ai/dsh-subagent-codex@0.1.5-alpha.1` (adopted generation,
revision `5dda764ed3aa172535a7967b06ff95d9cbfe536a`)

### What happens

`startCodexRun` forwards the `codex app-server --stdio` child's stderr straight
to the host process's stderr descriptor:

```js
const onStderr = (chunk) => {
  const bytes = typeof chunk === "string" ? Buffer.from(chunk) : chunk;
  try {
    writeFileSync(process.stderr.fd, bytes);
  } catch {}
};
child.stderr?.on("data", onStderr);
```

This is unconditional and lives for the whole run. On an interactive launch,
descriptor 2 is the terminal the frontend is drawing on.

### Why it breaks a terminal frontend

A TUI that redraws a bottom live region owns the screen: it tracks how many rows
the region occupies and where it left the cursor, and each redraw climbs that
geometry to erase the frame before drawing the next one. A write it did not
issue moves the cursor and scrolls the screen, so every later erase is measured
from the wrong origin. The frame's first rows are never erased again, scroll up
as ordinary output, and stay in native scrollback permanently — one more copy
per commit. In dshline this shows up as the root chrome (`╭─ dshline ─… ─╮`)
printed a second time, and the child's diagnostics printed *inside* the composer
frame. Reproduced against a real terminal emulator in
`packages/dshline/tests/foreign-writes.spec.ts`.

It also bypasses every mitigation a frontend could apply. Replacing or wrapping
`process.stderr.write` does not intercept it, because the forward never touches
the stream — it resolves `process.stderr.fd` on each write. The only local
containment is to substitute that descriptor, which is what dshline now does,
and which is not something a frontend should have to do to a process global.

Note this is not merely a cosmetic problem for the diagnostics either: the bytes
land on top of the frame, so they are unreadable in exactly the situation they
were forwarded to be read in.

### Preferred fix

Route the child's stderr through a Host-owned seam instead of the host's
terminal descriptor. Any of these retires the frontend shim:

1. **A Host diagnostic/logging sink.** The backend already has a diagnostic
   concept — `CodexAppServerWire` collects provider-authored diagnostic text and
   returns it on `CodexRunFailure` — so child stderr belongs on the same path
   rather than on a second, unowned one. A frontend then renders it where it can
   (dshline would commit it to scrollback above the live region, which is
   already what `Screen.commit` is for).
2. **A session/lifecycle event** carrying the run's diagnostic text, scoped to
   the delegating parent like the other `subagent/*` edges. Reaches every
   frontend, including non-terminal ones, through machinery that already exists.
3. **At minimum, gate it.** If raw forwarding must stay for local debugging, put
   it behind explicit configuration (a `forwardChildStderr` flag or an env
   variable) that defaults to off, and write through `process.stderr.write`
   rather than `writeFileSync(process.stderr.fd, …)` so a host that owns the
   terminal can intercept it.

### Why it should not be the frontend's problem

`process.stderr` is process-global state a plugin does not own. Today exactly
one package in the installed tree writes to a raw descriptor
(`grep -rn "writeFileSync(process" node_modules/@deepseek-ai/*/lib/*.js` → one
hit), so the invariant "Host diagnostics go through a Host seam" holds
everywhere else and is cheap to keep holding here.

### Also worth checking upstream

The same forward is timing-sensitive where the host's stderr is a **file**
rather than a terminal: Node backs a file-redirected `process.stderr` with
`SyncWriteStream`, which resolves `this.fd` on every write. Anything that
substitutes that descriptor — including a host doing so defensively — silently
captures ordinary stream writes too. Pinned in
`packages/dshline/tests/stderr-descriptor.spec.ts`.

